PR #2, opened 2020-12-01
[permissions] Move connection selector D-Bus interface to Internet. JB#52188

rainemak:


Tomin1 (approved):
Looks okay, I could use some more explanation why Ofono is needed in Internet though.

Also quite inconsistent use of Signed-off-by.

Tomin1 (approved):

PR #3, opened 2020-12-02
[permissions] Add a missing resource to Accounts. JB#52197

rainemak:


Tomin1 (approved):

Tomin1:
One thing, are you sure that the bug id is correct?

rainemak:
> One thing, are you sure that the bug id is correct?

Yes I am but good that you checked.

PR #5, opened 2020-12-02
[permissions] Improve Phone.permission. Contributes to JB#51739

Tomin1:
Add missing interfaces to Phone.permission and add some legacy things to
voicecall-ui.profile.


rainemak (approved):
Didn't test on device but the change itself looks good. Hence, approving. Let's see can we do something the profile file.

PR #8, opened 2020-12-02
[permissions] Include SSL configurations in Internet permission. JB#52191

abranson:


permissions/Internet.permission
```diff
@@ -12,6 +12,9 @@ protocol inet,inet6,netlink
 #net eth0
 #netfilter
 
+# Crypto config for SSL
+private-etc ssl,hosts,pki,crypto-policies
```
rainemak, line 16: These could be now removed from the Mozilla.permission.

rainemak (approved):
Feel free to remove the same line from the Mozilla.permissions. Approving already.

PR #9, opened 2020-12-03
[permissions] Use privileged-data option. Fixes JB#51951

spiiroin:
Privileged.permission file makes empty privileged data directory
available in sandbox. And FEATURE.permission files then make also
relevant privileged data subdirectories available.


rpm/sailjail-permissions.spec
```diff
@@ -10,6 +10,8 @@ BuildRequires: qt5-qmake
 BuildRequires: qt5-qttools-linguist
 BuildRequires: python3-base
 
+# TBD Conflicts: firejail < 0.9.63+git3
```
spiiroin, line 13: This firejail dep still needs to be updated.

Btw, I'm assuming that the lack of Requires: firejail is intentional.

spiiroin: After consultation: not needed as none of this has yet been in any release.

Tomin1 (approved):
There is still comment about conflicts with firejail in commit message. Other than that LGTM!

spiiroin:
> There is still comment about conflicts with firejail in commit message

Commit message edited.

spiiroin:
Rebased

PR #13, opened 2020-12-11
Add libaccounts-glib dbus entries to Accounts

llewelld:
Requires https://git.sailfishos.org/mer-core/libaccounts-glib/merge_requests/9

### DIFF
--- a/permissions/Accounts.permission
+++ b/permissions/Accounts.permission
@@ -21,6 +21,10 @@ whitelist ${HOME}/.local/share/accounts
 dbus-user.talk com.google.code.AccountsSSO.SingleSignOn
 dbus-user.broadcast com.google.code.AccountsSSO.SingleSignOn=*@/*
 
+# libaccounts-glib/libaccounts-qt
+dbus-user.own com.google.code.AccountsSSO.Accounts.*
+dbus-user.broadcast com.google.code.AccountsSSO.Accounts.*=com.google.code.AccountsSSO.Accounts@/ServiceType/*
+
 # buteo profiles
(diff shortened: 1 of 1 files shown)


llewelld:
Could you take a look please @Tomin1?

permissions/Accounts.permission
```diff
@@ -21,6 +21,10 @@ whitelist ${HOME}/.local/share/accounts
 dbus-user.talk com.google.code.AccountsSSO.SingleSignOn
 dbus-user.broadcast com.google.code.AccountsSSO.SingleSignOn=*@/*
 
+# libaccounts-glib/libaccounts-qt
+dbus-user.own com.google.code.AccountsSSO.Accounts.*
```
Tomin1, line 25: Is this needed? If this is just about being able to listen these com.google.code.AccountsSSO.Accounts.uuid_* names then, this shouldn't be needed as this is about owning those names (i.e. reserving a bus name for the sandboxed application itself).

llewelld: Yeah, the library is both a signal listener and an emitter, so it needs this to be able to own a name to send the signals out on (at least that's what I understood from our discussion: it doesn't need the name to send the signals, but it does need it to allow the signals to be filtered).

All of the instances basically store their data in the same central database, but they signal to each other that they've made a change over dbus in an attempt to keep all the instances synchronised. It's all done this way because there's no central daemon with a single view of the world.

Tomin1 (approved):
Alright then. Thanks for the explanation. This change works for me and looks fine.

PR #20, opened 2020-12-15
[permissions] Add jolla-settings to Phone permissions. Fixes JB#52483

pvuorela:
Phone app can show about page with *#07# and needs to call
settings d-bus for that.

@rainemak @Tomin1

### DIFF
--- a/permissions/Phone.permission
+++ b/permissions/Phone.permission
@@ -45,3 +45,7 @@ include /etc/sailjail/permissions/Audio.permission
 # Allow requesting sim pin
 dbus-user.talk com.jolla.PinQuery
 dbus-user.broadcast com.jolla.PinQuery=com.jolla.PinQuery.*@/*
+
+# BEG sessionbus-com.jolla.settings
+dbus-user.call com.jolla.settings=com.jolla.settings.ui.*@/*
+# END sessionbus-com.jolla.settings


pvuorela:
Or would it better just to have "talk" type of access as a simpler rule?

permissions/Phone.permission
```diff
@@ -38,3 +38,7 @@ dbus-system.talk org.ofono.SmartMessagingAgent
 # BEG sessionbus-com.jolla.voicecall.ui.resource
 dbus-user.talk com.jolla.voicecall.ui
 # END sessionbus-com.jolla.voicecall.ui.resource
+
+# BEG sessionbus-com.jolla.settings
+dbus-user.call com.jolla.settings=com.jolla.settings.ui.*@/*
```
rainemak, line 43: Can this be triggered by any other process than voicecall-ui-jolla?

As this is quite specific for phone app, I am thinking that this should go to voicecall-ui.profile instead.

pvuorela: Well the exact case I was here after is implemented in voicecall-jolla so in a way having the rule in .profile could indeed make sense. 

But actually there seems to be also a somewhat similar multisim settings page open call in components-telephony -> SimSelector and the inherited SimPicker. And that is used also in Sailfish.Messages and Sailfish.Contacts and don't know how far it spreads. 

So all in all not fully sure. Would it make sense if I move this into the profile now and scope this bug/fix to handle the special phone number, and the multisim case to be handled separately, somehow?

rainemak:
Rebase and merge... Approving.

rainemak (approved):
Rebase and merge

PR #26, opened 2020-12-17
[permissions] Add OTA provisioning interface to voicecall. Fixes JB#52523

pvuorela:
@rainemak @Tomin1


pvuorela:
Was assuming first the broadcast would suffice but required also talk to get that working.

Tomin1 (approved):

PR #27, opened 2020-12-18
Add permission and metadata descriptions to README

Tomin1:


README.md
```diff
@@ -1,4 +1,99 @@
 Sailjail permissions
 ====================
-
 Permissions for [Sailjail](https://github.com/sailfishos/sailjail/).
+These are [firejail](https://firejail.wordpress.com/) profile files with
```
Tomin1, line 4: Should this link to the webpage, upstream repository or our packaging of firejail?

README.md
```diff
@@ -1,4 +1,99 @@
...
+Sailjail specific metadata format.
```
Tomin1, line 5: Could be also Sailfish specific as Sailjail itself doesn't really use it for anything. It's used only on UI.

README.md
spiiroin: As long as upstream is identified and can be found, I think it would be better to refer to repos that we use for sailfish.

README.md
spiiroin: +1 for using "sailfish" rather than "sailjail"

README.md
```diff
@@ -1,4 +1,99 @@
...
+  - Access to Pictures directory and thumbnails.
+- Sharing
+  - Use of sharing pages in the application to share data via Bluetooth, email
+    or other accounts.
+- SyncFW
+  - Access to synchronization framework.
+- Videos
+  - Access to Videos directory and thumbnails.
+
+Internal permissions that applications generally should not use directly:
```
spiiroin, line 62: Also, some of these are mere dummies / placeholders that are not used even internally, e.g. FingerprintSensor.

Should we identify them separately? Or perhaps remove altogether.

README.md
```diff
@@ -1,4 +1,99 @@
...
+- Documents
+  - Access to Documents directory.
+- Downloads
+  - Access to Downloads directory.
+- Email
+  - Reading and sending emails. Access to email attachments.
+- Internet
+  - Using data connection and connecting to internet.
+- Location
+  - Use GPS and low accuracy positioning.
```
spiiroin, line 40: Is "low accuracy positioning" a common term?

README.md
```diff
@@ -1,4 +1,99 @@
...
+- MediaIndexing
+  - Access to Tracker to list files on device. If you have access to data
+    directory, you may want to use also this.
+- Messages
+  - Access to message data and to send SMS messages.
+- Mozilla
+  - If you use Gecko based WebView you need this.
```
spiiroin, line 47: I thought this got renamed. But apparently it was just the translated description.

Kind of separate issue from this PR, but these filenames do become a part of interface -> they should be as correct as possible as changing becomes difficult after they are in active use.

And there is similar situation with few other files too: Connman, Location, SyncFW

rainemak: Yeap, similar

README.md
rainemak: I'll remove "low accuracy"

README.md
rainemak: Doesn't harm much but I do agree that some are more like placeholders.

rainemak (approved):
Fine for me now. Needs @spiiroin approval as well.

spiiroin (approved):

PR #28, opened 2020-12-30
[permissions] Add contacts import wizard permission. Fixes JB#52582

mkenttala:
…52582

### DIFF
--- a/permissions/Contacts.permission
+++ b/permissions/Contacts.permission
@@ -16,6 +16,9 @@ include /etc/sailjail/permissions/Phone.permission
 # To launch Messages from contact card
 dbus-user.talk org.sailfishos.Messages
 
+# To launch contacts import wizard
+dbus-user.talk com.jolla.contacts.ui
+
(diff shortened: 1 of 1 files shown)


permissions/Phone.permission
```diff
@@ -49,3 +49,6 @@ dbus-user.broadcast com.jolla.PinQuery=com.jolla.PinQuery.*@/*
 # BEG sessionbus-com.jolla.settings
 dbus-user.call com.jolla.settings=com.jolla.settings.ui.*@/*
 # END sessionbus-com.jolla.settings
+
+# Contacts service
+dbus-user.talk com.jolla.contacts.ui
```
Tomin1, line 54: I wonder if this should be in Contacts.permission or even in voicecall-ui.profile.

rainemak: Yeap, Contacts.permission file feels more logical home for this.

rainemak (approved):

PR #33, opened 2021-01-11
[sailjail-permissions] Add office profile to allow SQlite storage.

dcaliste:
Sailfish-office is using a QML local storage to store last open position for documents. The path where this QML storage is located is not allowed by sailjail at the moment. This MR correct it.

@pvuorela,  you may want to standardize the location, using `sailfishos.org` or something similar instead of `Sailfish`. We need an additional MR in sailfish-office for this. What do you think ?

### DIFF
--- a/permissions/sailfish-office.profile
+++ b/permissions/sailfish-office.profile
@@ -1,0 +1,13 @@
+# -*- mode: sh -*-
+
+# Firejail profile for /usr/bin/sailfish-office
+
+# x-sailjail-translation-catalog =
+# x-sailjail-translation-key-description =
+# x-sailjail-description = Execute sailfish-office application
+# x-sailjail-translation-key-long-description =
+# x-sailjail-long-description =
+
+### APPLICATION
+mkdir     ${HOME}/.local/share/Sailfish/Sailfish Office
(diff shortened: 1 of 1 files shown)


pvuorela:
Nicely spotted! Created JB#52708 for tracking this.

Think there was a standardized path already allowed and on some places the organization and app names have been adjusted for that. And migration script for moving the files, though perhaps could also live if an upgrade lost pdf file positions one time.

@Tomin1 @rainemak what do you thnk?

Tomin1:
Regarding the paths, it should be indeed .local/share/org.sailfishos/Office and that doesn't require any changes to sailjail-permission. I seems that I had missed that when I created https://github.com/sailfishos/sailfish-office/pull/169.

See also:
- [Enable sandboxing for an application](https://github.com/sailfishos/sailjail-permissions#enable-sandboxing-for-an-application)
- https://github.com/sailfishos/sailfish-office/blob/master/sailfish-office.desktop#L15

dcaliste:
@tomin1 thanks for the pointers. So I'm closing this MR and I'm going to open one in sailfish-office to change organisation name and application name there.

PR #34, opened 2021-01-12
[permissions] Allow r/w access to user applications dir for sailfish-browser. Fixes JB#52554

NotKit:


permissions/sailfish-browser.profile
```diff
@@ -36,3 +36,8 @@ whitelist ${HOME}/.local/share/org.sailfishos/sailfish-browser
 
 mkdir     ${HOME}/.mozilla/captiveportal
 whitelist ${HOME}/.mozilla/captiveportal
+
+# Saving bookmarks to launcher feature
+# Override read-only from whitelist-common.inc
+mkdir      ${HOME}/.local/share/applications
```
abranson, line 42: Is the mkdir necessary? I'm concerned that it might end up with the wrong permissions. It's already been whitelisted earlier too.

NotKit: I was concerned that whilelisting would have no effect if path did not exist prior to application start, but I checked on another device and it seems to be there by default.

rainemak: Other way around creating this for the browser in the sailfish-browser.profile should not be an issue of any kind.

abranson: The mkdir should happen before the whitelist though, no? That can't happen in this instance, as it's in the whitelist-common.inc, so I'm not sure it should be happening at all.

abranson (approved):

PR #35, opened 2021-01-15
[sailjail-permissions] Add config file to Calendar permission. Contributes to JB#51280

chriadam:
The calendar application can make use of a configuration file which
stores e.g. calendar colour information.

### DIFF
--- a/permissions/Calendar.permission
+++ b/permissions/Calendar.permission
@@ -13,3 +13,7 @@ privileged-data Calendar
 
 mkdir     ${PRIVILEGED}/autofill
 privileged-data autofill
+
+mkdir     ${HOME}/.config/nemo
+whitelist ${HOME}/.config/nemo
+whitelist ${HOME}/.config/nemo/nemo-qml-plugin-calendar.conf


permissions/Calendar.permission
```diff
@@ -13,3 +13,7 @@ privileged-data Calendar
 
 mkdir     ${PRIVILEGED}/autofill
 privileged-data autofill
+
+mkdir     ${HOME}/.config/nemo
+whitelist ${HOME}/.config/nemo
```
chriadam, line 18: This is suboptimal, as I only want the Calendar app to be able to access the specific file: ${HOME}/.config/nemo/nemo-qml-plugin-calendar.conf

But that directory may or may not include other unrelated config files also...

Any hints?

Tomin1: The only thing I can think of would be to move that file to a subdirectory and whitelist that.

abranson: I think we're going for a /owner/app structure in there too. I already see a com.jolla/calendar directory. Could it be moved to there?

chriadam: Actually, discussed with Pekka that after my PR is merged, we don't actually use the conf file for anything.  Since the calendar widget gets loaded into the homescreen we can rely on it to do the migration, and then we don't need this permission change at all.

chriadam:
Declining, not required since the conf file shouldn't be used any more, after https://git.sailfishos.org/mer-core/nemo-qml-plugin-calendar/merge_requests/72/diffs

PR #36, opened 2021-01-15
[permissions] Add permissions for qr service. Contributes to JB#49377

HandsomeJaack:


permissions/Qr.permission
```diff
@@ -0,0 +1,11 @@
+# -*- mode: sh -*-
+
+# x-sailjail-translation-catalog = sailjail-permissions
+# x-sailjail-translation-key-description = permission-la-qr
+# x-sailjail-description = QR
+# x-sailjail-translation-key-long-description = permission-la-qr_description
+# x-sailjail-long-description = Recognize QR codes
```
rainemak, line 7: Recognize QR codes... QR code reading

okodron: Indeed

rainemak: On Android camera permissions it is "Allow QR code reader" but that doesn't of the version and provider. We already dropped those "Allow" prefixes so please do not add that.

So my vote goes to "QR code reader" (not reading)

rainemak: See above ^

permissions/Qr.permission
```diff
@@ -0,0 +1,11 @@
+# -*- mode: sh -*-
+
+# x-sailjail-translation-catalog = sailjail-permissions
+# x-sailjail-translation-key-description = permission-la-qr
+# x-sailjail-description = QR
```
rainemak, line 5: Pondering about this short description... Somehow just "QR" sounds odd.

Tomin1: I would maybe use these:
descripiton: QR codes
long-description: Read QR codes

okodron (approved):

permissions/Camera.permission
```diff
@@ -11,3 +11,7 @@
 dbus-system.talk org.maemo.resource.manager
 dbus-system.broadcast org.maemo.resource.manager=org.maemo.resource.manager.*@/*
 # END systembus-org.maemo.resource.manager
+
+# BEG sessionbus-org.sailfishos.zxing
+dbus-user.talk org.sailfishos.zxing
```
rainemak, line 16: No need to listen signal from "org.sailfishos.zxing"?

HandsomeJaack: Works fine without it.

rainemak: Ok, and if there are issues later it's easy to add "dbus-user.broadcast" rules for this as well.

rainemak (approved):

PR #40, opened 2021-01-20
Add a profile to allow GnuPG crypto operations

### LINKED ISSUE #15: Access to Secret daemon is missing
I would like to add a profile to give access to the Secret daemon. The use case is the email signature process:
* compose an email with signature, and select send,
* the `gpgme` library (see QMF plugin for cryptography) will end up calling the `/usr/bin/pinentry` to ask for the passphrase of the signing key (this is done internally by contacting the gpg-agent on a socket,
* the pinentry program (see gnupg plugin in sailfish-secret) will converse with the Secret daemon to inquire the user for its passphrase or look at the cached value in secret storage. This is done through a peer-to-peer DBus connexion via a socket. It will require access to the secret database storage also.
* the pinentry will send back the passphrase to gpgme so the signature can proceed.

What should be granted as permissions to allow this ? @rainemak, @tomin or @pvuorela if you may give some insights on the matter please?

dcaliste:
Fixes #15.

@spiiroin and @pvuorela this would allow to use crypto operations in jolla-email. It still requires to add the GnuPG permission to the jolla-email desktop file though.

Before the patch, you cannot sign an outgoing email.

### DIFF
--- a/permissions/Base.permission
+++ b/permissions/Base.permission
@@ -18,6 +18,9 @@ writable-run-user
 # we use "private-etc selinux" -> must not blacklist it
 noblacklist /etc/selinux
 
+# we need to be able to whitelist it later, while it's in disable-common.inc
+noblacklist ${HOME}/.gnupg
+
 # Persistent local customizations
 # Persistent global definitions
 #include globals.local
--- a/permissions/Email.permission
+++ b/permissions/Email.permission
@@ -44,6 +44,9 @@ include /etc/sailjail/permissions/SyncFW.permission
 # Media Indexing for above
 include /etc/sailjail/permissions/MediaIndexing.permission
 
+# Sign emails with GnuPG
+include /etc/sailjail/permissions/GnuPG.permission
+
 # Global address list
 dbus-user.talk org.sailfishos.easdaemon
 dbus-user.broadcast org.sailfishos.easdaemon=org.sailfishos.easdaemon.*@/*
--- a/permissions/GnuPG.permission
+++ b/permissions/GnuPG.permission
@@ -1,0 +1,14 @@
+# -*- mode: sh -*-
+
+# x-sailjail-translation-catalog =
+# x-sailjail-translation-key-description =
+# x-sailjail-description = GnuPG
+# x-sailjail-translation-key-long-description =
+# x-sailjail-long-description = Use cryptographic operations with GnuPG
+
+mkdir     ${HOME}/.gnupg
+whitelist ${HOME}/.gnupg
+
+private-bin gpg2
+private-bin gpg-agent
+private-bin pinentry
--- a/permissions/jolla-email.profile
+++ b/permissions/jolla-email.profile
@@ -18,6 +18,7 @@
 # x-sailjail-permission = Internet
 # x-sailjail-permission = Connman
 # x-sailjail-permission = Calendar
+# x-sailjail-permission = GnuPG
 # for starting messageserver5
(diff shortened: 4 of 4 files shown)


permissions/sailfish-office.profile
```diff
@@ -0,0 +1,13 @@
...
+
+# x-sailjail-translation-catalog =
+# x-sailjail-translation-key-description =
+# x-sailjail-description = Execute sailfish-office application
+# x-sailjail-translation-key-long-description =
+# x-sailjail-long-description =
+
+### APPLICATION
+mkdir     ${HOME}/.local/share/Sailfish/Sailfish Office
+whitelist ${HOME}/.local/share/Sailfish/Sailfish Office
```
pvuorela, line 13: Not gpg nor needed anymore?

pvuorela:
cc: @rainemak and @Tomin1 

Wonder should we include the privileges in email rather than introducing a new permission. People might get confused if email wanted gpg but don't include the functionality there by default.

permissions/sailfish-office.profile
dcaliste: Oups, an old local modification that I commited by error with a `git commit -a`.

dcaliste: Or worst, an old commit on top... Removing it.

permissions/Base.permission
```diff
@@ -18,6 +18,9 @@ writable-run-user
 # we use "private-etc selinux" -> must not blacklist it
 noblacklist /etc/selinux
 
+# we need to be able to whitelist it later, while it's in disable-common.inc
+noblacklist ${HOME}/.gnupg
```
spiiroin, line 22: Need to double-check, but ...

When "blacklist", "noblacklist", "whitelist" etc directives are read from config files, they are just queued for later processing - in fixed per directive order -> it should be so that the .gnupg directory can remain blacklisted by default, and both "noblacklist" and "whitelist" can be in feature/app specific config file -> depending on what the intent is, "noblacklist" can/should be moved to GnuPG.permission in this case.

spiiroin (changes requested):

permissions/Base.permission
dcaliste: It seems to me that a blacklist order when arriving before a noblacklist is taking precedence. I mean, we cannot revert a blacklisted entry after it is done. I cannot find the documentation for that again... The `/etc/firejail/gpg.profile` is doing it the same way : first you noblacklist to cancel later blacklisting in including `disable-common.inc`.

I'm quite sure, I tested to put the `noblacklist` in the GnuPG.permission file, without touching the Base one, but without any success.

Tomin1:
> cc: @rainemak and @Tomin1
> 
> Wonder should we include the privileges in email rather than introducing a new permission. People might get confused if email app wanted gpg but don't include the functionality there by default.

I think this sounds like a good idea, at least at this point. It could be still worth it to keep this in a separate file (possibly without translation strings) and then include it from Email.permission.

dcaliste:
> It could be still worth it to keep this in a separate file (possibly without translation strings) and then include it from Email.permission.

I've rebased on master and pushed the modifications you suggested.

Tomin1 (approved):
LGTM

I don't know if @spiiroin still wants something regarding that noblacklist but AFAICS it looks like reasonable compromise to use it here.

pvuorela:
No objections lately, I assume this can be merged.

PR #41, opened 2021-01-21
Allow Camera to access sensors and device info for Exif.

mlehtima:


permissions/Base.permission
```diff
@@ -141,6 +141,10 @@ protocol unix
 ### environment
 shell none
 
+### device info
+private-etc hw-release
+whitelist /usr/share/ssu/board-mappings.d
```
spiiroin, line 146: Hmm, release and ssu stuff... not directly related to this change, but

We already allow /etc/os-release, but as that is a symlink to /etc/sailfish-release -> is it actually working? should both names be whitelisted.

Also we have those HW feature probes that modify ui behavior in select places. Probably it is just settings, but we might need allow access to /usr/share/csd/settings.d now/later.

And then there is ssu.ini that contains public consumption data like "brand" among not so public things.

permissions/Sensors.permission
```diff
@@ -5,3 +5,7 @@
 # x-sailjail-description = Sensors
 # x-sailjail-translation-key-long-description =
 # x-sailjail-long-description =
+
+dbus-system.talk com.nokia.SensorService
+
+whitelist /var/run/
```
spiiroin, line 11: If this is for /var/run/sensord.sock access, there ought to be no reason to whitelist the containing directory, or?

mlehtima: Whitelisting just the socket is not possible because of limitations in firejail.

mlehtima: The socket was moved /run/sensord.sock which is automatically allowed and this line was removed.

permissions/Base.permission
mlehtima: This depends on which parts of the device info API we want to allow, the qml side is quite simple but the c++ side has more features.

Tomin1: I would now do the things that are useful for fixing camera and file another task for the rest. Also I think the changes here look good enough.

Tomin1 (approved):

permissions/Base.permission
rainemak: Let's do so... New tasks should be created from the rest.

PR #42, opened 2021-01-22
[sailjail-permissions] Add Keys dir to Accounts. Contributes to JB#52867

pvuorela:
The keyprovider system used by some gallery plugins has user written
configuration here. Could be good to allow even if the static
data is commonly used.

@Tomin1 @rainemak


pvuorela:
Moved to Accounts as it has the static data file already.

Tomin1 (approved):
LGTM, but you should still edit the PR title to match the commit line.

PR #43, opened 2021-01-29
[qr] Enable QR reading in lockscreen. Contributes to JB#52866

HandsomeJaack:


pvuorela:
I don't think this belongs to camera permissions. If no better then a .profile file could be an option but would hope that wouldn't be necessary.

HandsomeJaack:
The modification is indeed unnecessary, because the expected behavior was achieved without it.

PR #46, opened 2021-02-03
[permissions] Add Qt's logging rules. Contributes to JB#52717

Tomin1:


permissions/Base.permission
```diff
@@ -166,3 +166,6 @@ mkdir ${PRIVILEGED}
 whitelist ${PRIVILEGED}
 # but by default all content is hidden
 privileged-data -
+
+### Qt's logging rules
+whitelist ${HOME}/.config/QtProject/qtlogging.ini
```
Tomin1, line 171: Do we need to make this read only somehow or is this just copied to the sandbox when setting it up?

rainemak: Before this...
mkdir ${HOME}/.config/QtProject
and...
read-only ${HOME}/.config/QtProject/qtlogging.ini

Does read-only directive work if the target file doesnt exist?

Tomin1: I didn't add mkdir because I think we don't need to have that file in the sandbox unless it exists when the sandbox is created. The rules are anyway applied when the application is launched and changes afterwards don't affect it.

> Does read-only directive work if the target file doesnt exist?

I think it does. It also doesn't seem to require that the directory exists, so I could just add that read-only like that. In case the file doesn't exist, whitelisting just ignores it and if the application creates the file the changes won't be persistent which would again be fine as we don't want sandboxed apps to change the rules for other applications.

This is why I was wondering about that whitelist vs. copying the file:
>        whitelist file_or_directory
>              Whitelist  directory or file. A temporary file system is mounted
>              on the top directory, and the whitelisted files are mount-binded
>              inside.  Modifications  to whitelisted files are persistent, ev‐
>              erything else is discarded when the sandbox is closed.  The  top
>              directory  could  be  user home, /dev, /etc, /media, /mnt, /opt,
>              /srv, /sys/module, /usr/share, /var, and /tmp.
https://firejail.wordpress.com/features-3/man-firejail-profile/
I don't really like the sound of that mount-binded part because it causes this on older kernels:
> [defaultuser@JollaC ~]$ mv .config/QtProject/qtlogging.ini .config/QtProject/qtlogging.ini.rmd
> mv: can't rename '.config/QtProject/qtlogging.ini': Device or resource busy

If it were copied instead, this would be avoided. Also if the application then messes with the file, it doesn't really matter because it wouldn't affect other applications.

rainemak (approved):

Tomin1:
Is there any way we could avoid locking that file?

Tomin1:
> Is there any way we could avoid locking that file?

Well, there haven't been any better ideas. We probably need to consider some improvements to firejail at some point but let's go with this for now at least.

PR #50, opened 2021-02-25
[permissions] Add Contacts dbus-user permissions for qtcontacts-sqlit…

blammit:
…e. Contributes to JB#53239

### DIFF
--- a/permissions/Contacts.permission
+++ b/permissions/Contacts.permission
@@ -19,6 +19,9 @@ dbus-user.talk org.sailfishos.Messages
 # To launch contacts import wizard
 dbus-user.talk com.jolla.contacts.ui
 
+dbus-user.own org.nemomobile.contacts.sqlite.*
+dbus-user.broadcast *=org.nemomobile.contacts.sqlite.*@/org/nemomobile/contacts/sqlite
+
 mkdir     ${HOME}/.local/share/system/Contacts
 whitelist ${HOME}/.local/share/system/Contacts
 


Tomin1 (approved):

permissions/Contacts.permission
```diff
@@ -19,6 +19,9 @@ dbus-user.talk org.sailfishos.Messages
 # To launch contacts import wizard
 dbus-user.talk com.jolla.contacts.ui
 
+dbus-user.own org.nemomobile.contacts.sqlite.*
+dbus-user.broadcast org.nemomobile.contacts.sqlite.*=org.nemomobile.contacts.sqlite.*@/org/nemomobile/contacts/sqlite
```
pvuorela, line 23: Just wondering would it allow skipping the service registration in app side if this was *=org.nemomobile...

Tomin1: I didn't check it again but IIRC no, it doesn't work like that. Currently it's just not possible to allow signals from unique bus names.

pvuorela: FWIW such a format is already used on Base.permission for standard interfaces.

Tomin1: Funny. I think it can be used as allow from any well-known bus name but unique bus names are still not allowed in the code.

pvuorela: Anyhow, could think *=... being more correct here as it should be possible to trigger the signal regardless of the registered services.

blammit: Updated to use *=...
Verified that the app side still requires the service registration with this change.

PR #53, opened 2021-03-09
[permissions] Move noblacklist to correct file. Contributes to JB#53456

Tomin1:


dcaliste:
I may be wrong, but when I tested it, it seems to me that putting the `noblacklist` keyword after the `blacklist` one as mentioned in the comment, was not working. It seems from the man page https://man7.org/linux/man-pages/man5/firejail-profile.5.html that the `noblacklist` keyword is working for `blacklist` coming _after_.

I need to test it again to see if I'm wrong.

dcaliste:
@Tomin1 I missed the other MR moving the Base permissions at the end. My bad.

LGTM.

spiiroin:
LGTM

PR #54, opened 2021-03-17
Add script to visualise permission relationships.

Tomin1:
This script allows to visualise permission relationships. It mostly
deals with includes of other permissions. This is useful to tell if
there are problematic implicit dependencies.

It also shows if permission can be directly included or if it's so
called internal permission. That heuristic is based on missing
translation category which is mostly correct indicator (there may be
false negatives).

Additionally it shows if permission defines privileged data directory.

Graphviz is required to use this tool.

### DIFF
--- a/tools/dependencies.py
+++ b/tools/dependencies.py
@@ -1,0 +1,182 @@
+#!/usr/bin/env python3
+#
+# Visualises permission dependencies
+# Copyright (c) Jolla Ltd. 2021
+#
+# You may use this file under the terms of the BSD license as follows:
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#
+#   1. Redistributions of source code must retain the above copyright
+#      notice, this list of conditions and the following disclaimer.
+#   2. Redistributions in binary form must reproduce the above copyright
+#      notice, this list of conditions and the following disclaimer
+#      in the documentation and/or other materials provided with the
+#      distribution.
+#   3. Neither the names of the copyright holders nor the names of its
+#      contributors may be used to endorse or promote products derived
+#      from this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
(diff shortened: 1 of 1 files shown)


tools/dependencies.py
```diff
@@ -0,0 +1,184 @@
...
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+# Requries graphviz library
+# Code: https://github.com/xflr6/graphviz
+# PyPI: https://pypi.org/project/graphviz/
+# It may also be available in your distribution's repositories
+import graphviz
+
+import argparse
+import os
```
Thaodan, line 41: That looks redundant.

Tomin1: Yeah, some left over. I'll remove it, thanks.

Thaodan (approved):

tools/dependencies.py
```diff
@@ -0,0 +1,183 @@
...
+    forest = Forest.build(input_directory)
+    if debug:
+        print(forest)
+        for node in forest:
+            print(node)
+    digraph = forest.digraph()
+    digraph.attr(pack='true')
+    unflatten = digraph.unflatten(stagger=2)
+    if output_file is None:
+        with tempfile.TemporaryDirectory() as tmpdir:
```
Tomin1, line 153: This shouldn't be here. I forgot to remove it after changing this to use mktemp().

spiiroin (approved):
LGTM

tools/dependencies.py
```diff
@@ -0,0 +1,182 @@
...
+                        if directories != "-":
+                            self.privileged = True
+                    else:
+                        match = Forest.TEXT.match(line)
+                        if match and match.group("text"):
+                            self.text = match.group("text")
+                        elif Forest.CATALOG.match(line):
+                            self.user_visible = True
+
+            forest.nodes[path.name] = self
```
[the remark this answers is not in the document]
Tomin1: Well, yeah, circular dependency would break this (-> recursion error). Good catch. I should move this line up before parsing the file.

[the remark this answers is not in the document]
Tomin1: Debug output will still cause recursion error but that's a bit harder to solve. Maybe we don't need to care about that too much.

PR #60, opened 2021-04-01
Default to "net none"

spiiroin:
Now that maliit uses regular unix socket instead of an abstract one,
applications can be put into private net namespaces via "net none".
    
Which then needs to be undone for those applications that have been
granted internet access.

### DIFF
--- a/permissions/Base.permission
+++ b/permissions/Base.permission
@@ -132,11 +132,12 @@ dbus-user.broadcast com.nokia.time=com.nokia.time.*@/*
 # BEG sessionbus-org.maliit.server.resource
 dbus-user.talk org.maliit.server
 dbus-user.broadcast org.maliit.server=org.maliit.server.*@/*
+read-only ${RUNUSER}/maliit-server
 # END sessionbus-org.maliit.server.resource
 
 ### network
 protocol unix
-# net none <== can't be used as "net none" breaks maliit
+net none
 
 ### environment
 shell none
--- a/permissions/Internet.permission
+++ b/permissions/Internet.permission
@@ -8,6 +8,7 @@
 
 # XXX: does "netlink" really belong here?
 ### network
+ignore net none
 protocol inet,inet6,netlink
 #net eth0
 #netfilter


permissions/Base.permission
```diff
@@ -132,11 +132,12 @@ dbus-user.broadcast com.nokia.time=com.nokia.time.*@/*
 # BEG sessionbus-org.maliit.server.resource
 dbus-user.talk org.maliit.server
 dbus-user.broadcast org.maliit.server=org.maliit.server.*@/*
+whitelist ${RUNUSER}/maliit-server/dbus-socket
```
spiiroin, line 135: Started to wonder what happens if maliit ever gets restarted? Does sandboxed app see a stale socket? Would something like allowing read-only access to containing directory work better?

Tomin1: Tested these on device, if we whitelist the file, the keyboard doesn't reappear when maliit-server is restarted but then if I whitelist the directory it does. However it didn't stop maliit-server from working as keyboard was there once I restarted the app, unlike I anticipated on previous comment.

spiiroin: After taking a peek at firejail source and some testing:

    whitelist DIR
    read-only DIR

seems to be the combination with desired outcome: sandboxed application gets a "live" view to the real directory, but can't modify anything underneath the DIR -> have notes app running and vkbd open; restart maliit -> vkbd goes away, and after a while comes back up and works; but things like firejail --put=$(pidof jolla-notes) foo /run/user/100000/maliit-server/bar fail due to EROFS

I guess the same applies for similar situations like e.g. the pulse audio socket in this PR.

spiiroin:
The reason behind pulse audio woes seems to be: There are no whitelist rules for runuser. Adding the first one affecs all of runuser -> immediate wholesale visibility reconfiguration of all things in there would be needed.

As long as we do not add whitelist rules, read-only is sufficient for maliit-change and nothing needs to be changed regarding e.g. pulseaudio.

Review commits squashed.

Tomin1 (approved):

PR #61, opened 2021-04-12
Hide well-known directories by default

Tomin1:
Hide well-know directories by default. These rules mount an empty
read-only directory on top of well-known directories unless the
respective permissions have been granted. Update documentation about
these rules. Also mention that libsailfishapp will set OrganizationName
and ApplicationName correctly.


Tomin1:
@okodron, would you like to review this?

okodron (approved):

rainemak (approved):
Approving as well.

PR #62, opened 2021-04-13
Add user common dirs permissions

LaakkonenJussi:
Use includes to add permissions to `Downloads`, `Documents`, `Music`, `Pictures` and `Videos`. In addition include the common user directory `Public` but exclude `Templates` as unneeded and `Desktop` as not wanted our desktop files are handled differently.

Also fix a copy-paste error in `README.md`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ OrganizationName and ApplicationName is used for granting for the application wr
 Access above directories from the application through QStandardPaths
 1. Application data location - QStandardPaths::writableLocation(QStandardPaths::AppDataLocation)
 2. Application cache location - QStandardPaths::writableLocation(QStandardPaths::CacheLocation)
-3. Application config location  - QStandardPaths::writableLocation(QStandardPaths::CacheLocation)
+3. Application config location  - QStandardPaths::writableLocation(QStandardPaths::ConfigLocation)
 
 Only these directories can be used for storing application specific data that needs to persist over
 application restarts.
@@ -80,7 +80,7 @@ set in desktop file.
 
 ### Files shared with other applications
 
-Well-known directories such as Documents, Downloads, Music, Pictures and Videos contain files that
+Directories under user home such as Documents, Downloads, Music, Pictures and Videos contain files that
 are available for applications with respective permissions. Those directories must be used when the
 data needs to be accessible by other applications.
 
@@ -115,8 +115,10 @@ Permissions that applications may use (names are subject to change):
 | Music | Access to Music directory, playlists and coverart cache. |
 | Phone | Make Phone calls, either directly or through system voice call UI. |
 | Pictures | Access to Pictures directory and thumbnails. |
+| PublicDir | Access to Public directory. |
 | Sharing | Use of sharing pages in the application to share data via Bluetooth, email or other accounts. |
 | Synchronization | Access to synchronization framework. |
+| UserDirs | Access to Documents, Downloads, Music, Pictures, Public and Video directories. |
 | Videos | Access to Videos directory and thumbnails. |
 
 Internal permissions that applications generally should not use directly:
--- a/permissions/Base.permission
+++ b/permissions/Base.permission
@@ -169,10 +169,12 @@ privileged-data -
 whitelist ${HOME}/.config/QtProject/qtlogging.ini
 read-only ${HOME}/.config/QtProject/qtlogging.ini
 
-### Disable access to well-known directories
+### Disable access to well-known directories (XDG user directories) by default
 # These rules say that there is an empty read-only directory inside
 # sandbox. They are prevented separately for each directory with
 # noblacklist rule in their respective permission file
+mkdir     ${HOME}/Desktop
+blacklist ${HOME}/Desktop
 mkdir     ${HOME}/Documents
 blacklist ${HOME}/Documents
 mkdir     ${HOME}/Downloads
@@ -185,3 +187,7 @@ mkdir     ${HOME}/Music
 blacklist ${HOME}/Music
 mkdir     ${HOME}/Videos
 blacklist ${HOME}/Videos
+mkdir     ${HOME}/Public
+blacklist ${HOME}/Public
+mkdir     ${HOME}/Templates
+blacklist ${HOME}/Templates
--- a/permissions/UserDirs.permission
+++ b/permissions/UserDirs.permission
@@ -1,0 +1,20 @@
+# -*- mode: sh -*-
+
+# x-sailjail-translation-catalog = sailjail-permissions
+# x-sailjail-translation-key-description = permission-la-userdirs
+# x-sailjail-description = User directories
+# x-sailjail-translation-key-long-description = permission-la-userdirs_description
+# x-sailjail-long-description = Access common user directories
+
+include /etc/sailjail/permissions/Documents.permission
+
+include /etc/sailjail/permissions/Downloads.permission
+
+include /etc/sailjail/permissions/Music.permission
+
+include /etc/sailjail/permissions/Pictures.permission
+
+include /etc/sailjail/permissions/PublicDir.permission
+
+include /etc/sailjail/permissions/Videos.permission
+
--- a/permissions/PublicDir.permission
+++ b/permissions/PublicDir.permission
@@ -1,0 +1,11 @@
+# -*- mode: sh -*-
+
+# x-sailjail-translation-catalog = sailjail-permissions
+# x-sailjail-translation-key-description = permission-la-publicdir
+# x-sailjail-description = Public directory
+# x-sailjail-translation-key-long-description = permission-la-publicdir_description
+# x-sailjail-long-description = Access Public directory
+
+mkdir       ${HOME}/Public
+noblacklist ${HOME}/Public
+whitelist   ${HOME}/Public


permissions/UserDirs.permission
```diff
@@ -0,0 +1,24 @@
...
+
+include /etc/sailjail/permissions/Documents.permission
+
+include /etc/sailjail/permissions/Downloads.permission
+
+include /etc/sailjail/permissions/Music.permission
+
+include /etc/sailjail/permissions/Pictures.permission
+
+mkdir     ${HOME}/Public
```
Tomin1, line 17: While I'm not totally sure if we should have this directory as a permission, but perhaps, I think in any case it should be a separate permission that is then just included here. That way application can choose to have this permission without access to any of the other directories. The same applies to Templates directory.

LaakkonenJussi: Like 3730ce50bde007216149e30de9010c5f6fc14a3a ?

Tomin1: Yes

LaakkonenJussi: Ok, I'll squash that then to the main commit and wait for final verdict.

LaakkonenJussi: Done.

Tomin1: Oh, sorry, I forgot about the latest change to this: You need to add also noblacklist rule for both directories and then blacklist it in Base.permission ([see the comment at the end](https://github.com/sailfishos/sailjail-permissions/commit/76e8aa5cfb1b279d9a53f1c13842ebc57de812f4#diff-cbb6d3ce8d0bd38db5cb3a5cf84c67aa0c8a7130efb833a164de2613ec01cabfR171)). That way application sees read-only version of the directory unless it has the needed permission.

LaakkonenJussi: Added similar changes with 79cadf1 and updated `README.md` as well.

README.md
```diff
@@ -115,8 +115,11 @@ Permissions that applications may use (names are subject to change):
 | Music | Access to Music directory, playlists and coverart cache. |
 | Phone | Make Phone calls, either directly or through system voice call UI. |
 | Pictures | Access to Pictures directory and thumbnails. |
+| PublicDir | Access to Public directory. |
 | Sharing | Use of sharing pages in the application to share data via Bluetooth, email or other accounts. |
 | Synchronization | Access to synchronization framework. |
+| TemplatesDir | Access to Templates directory. |
+| UserDirs | Access to XDG user directories: Documents, Downloads, Music, Pictures, Public, and Templates (except Desktop that is unused in Sailfish OS). |
```
spiiroin, line 122: nit: Feel free to ignore. That "(except ..)" portion feels a bit incomplete or something. Maybe  "all except..." or "Desktop is not include because..." or something.

spiiroin (approved):

README.md
LaakkonenJussi: Yeah I was thinking if it is reasonable to mention as it is excluded in the list already. Maybe for clarity I just drop that ()-content.

LaakkonenJussi: Dropped the excess content with: 864cd19 - I guess it is better now.

Tomin1 (approved):
LGTM, remember to squash. :+1:

LaakkonenJussi:
@okodron could you review this as well?

permissions/TemplatesDir.permission
```diff
@@ -0,0 +1,11 @@
...
+
+# x-sailjail-translation-catalog = sailjail-permissions
+# x-sailjail-translation-key-description = permission-la-templatesdir
+# x-sailjail-description = Templates directory
+# x-sailjail-translation-key-long-description = permission-la-templatesdir_description
+# x-sailjail-long-description = Access Templates directory
+
+mkdir       ${HOME}/Templates
+noblacklist ${HOME}/Templates
+whitelist   ${HOME}/Templates
```
rainemak, line 11: I'd not like to promote "Templates" directory (same applies to Desktop which you don't have here). Public is somewhat reasonable.

rainemak: Not that harmful either but mkdir + blacklist in Base makes sense.

LaakkonenJussi: It is a part of standard Linux desktop. There is a reason to ignore `Desktop` as we handle desktop items differently. There is no real reason to exclude `Templates` especially since if some future partner/client expects it to be there as per to normal XDG user directories, e.g., [Arch ref](https://wiki.archlinux.org/index.php/XDG_user_directories).

permissions/Base.permission
```diff
@@ -185,3 +185,7 @@ mkdir     ${HOME}/Music
 blacklist ${HOME}/Music
 mkdir     ${HOME}/Videos
 blacklist ${HOME}/Videos
+mkdir     ${HOME}/Public
+blacklist ${HOME}/Public
+mkdir     ${HOME}/Templates
+blacklist ${HOME}/Templates
```
rainemak, line 191: Worth create and blacklist also Desktop. @Tomin1 ?

LaakkonenJussi: Makes sense unless we explicitly disallow every dir access by default? I've not been involved in this long enough to understand this completely yet..

permissions/TemplatesDir.permission
rainemak: Correct but intention is to simplify basic use case. If need surfaces later, simple to add.

LaakkonenJussi: Well this is included in the main `UserDirs` permission. I like to think ahead, not to my feet.

LaakkonenJussi: My reasoning here is that if you get the "well known dirs" permission - which effectively means the XDG user directories, you should get the access to all of them that are applicable in our OS.

LaakkonenJussi: And we do have `xdg-user-dir` tool, which [supports these](https://manpages.debian.org/stretch/xdg-user-dirs/xdg-user-dir.1.en.html).

LaakkonenJussi: Maybe in our case the DESKTOP should actually point in the config to the actual location of desktop files: `$HOME/.local/share/applications` ? So we could include `Desktop` as such and fully support `xdg-user-dir` tool with these permissions as well?

permissions/Base.permission
Tomin1: Note: We actually have a mention of Desktop in WebView.permission but I'm not sure it should be there and I think we should get rid of that. It's probably not used in practice but again one of those, it happens to access it but doesn't use it so it's there. If we blacklist it here and we keep it there, then there needs to be noblacklist rule.

That said, maybe we should blacklist it to discourage use. By default creating files to wrong places causes them disappear when the sandbox is destructed (application is closed) but this way you can't create files to some of the wrong places which gives developer (or user) immediate feedback instead of wondering what happened.

LaakkonenJussi: @Tomin1 or we change the behavior of `XDG_DESKTOP_DIR` to point to where the user app could create additional shortcuts? See https://github.com/sailfishos/sailjail-permissions/pull/62#discussion_r613094062

permissions/TemplatesDir.permission
Tomin1: That same reasoning, that we don't use Desktop directory so it's not needed here, could be applied to Templates as well. We don't use it for anything really and allowing it could make someone think that it's used somewhere. Having Public directory as permission is easier to reason about, it's kind of general place of sharing data between apps when nothing else applies.

I don't think xdg desktop should point to $HOME/.local/share/applications because that's a different thing. It's not desktop directory, it's place to put desktop files that are shown in menu. Although, IIRC, we don't use that either, only the system wide directory.

LaakkonenJussi: It is not a good argument that we do not use it but it exists in standard dirs. If you think a Linux application developer in general and maybe porting of an app to SFOS, then it may be that those dirs are used/required.

`$HOME/.local/share/applications` is not excluded in use but still it works fine and has always worked, I've seen some Harbour apps using it. And I've used it as well to add custom items to our "desktop" view.

LaakkonenJussi: It is just bad to give an assumption that with these permissions you get access to all normal dirs but except you don't. Lets not talk about "well known directories" then. Lets talk about Sailfish OS accepted user home resources?

rainemak: We are here defining from our point of "well known directories". The UserDirs permission do not even need to be bound to .config/user-dirs.dirs by any means. It's matter of the Permission itself and how do we define it.

permissions/Base.permission
rainemak: > Note: We actually have a mention of Desktop in WebView.permission but I'm not sure it should be there and I think we should get rid of that. It's probably not used in practice but again one of those, it happens to access it but doesn't use it so it's there. If we blacklist it here and we keep it there, then there needs to be noblacklist rule.

Having "Desktop" in the WebView.permission sounds wrong indeed.

permissions/TemplatesDir.permission
LaakkonenJussi: @rainemak Um no, simply NO! We are not defining something that is quite commonly accepted to be our own version of something else. I'm not going to be a part of it.

LaakkonenJussi: Lets put it this way, since, for example, `xdg-user-dir` tool is supplied by us it is kind of ridiculous that if application wants `UserDirs` permissions and then can just use a portion of the dirs that are available this tool. It confuses the developer, especially if the dirs are used to retrieve with the tool. One could expect that what that tool provides are usable with the same permissions, right? I see no counter argument to this than "we don't want to" and for me it is not enough.

We should also modify the `Desktop` entry as there is a real argument on not using that. Maybe the .config/user-dirs.dirs should be modified.

permissions/Base.permission
LaakkonenJussi: Unless we want to give access via that to the `$HOME/.local/share/applications` to add URL desktop entries there to be visible on _our_ desktop?

LaakkonenJussi: Eventually resulting in modifying `.config/user-dirs.dirs` as well.

rainemak: Correct and that's another task and not related to this per se.

LaakkonenJussi: Not related I agree but a worthy issue to raise here as well but not to venture into.

permissions/TemplatesDir.permission
LaakkonenJussi: On _normal_ Linux desktop one can put any document there to be used as a template for a new document to be created. Easy to try, just put a document there and use "Create document" anywhere on your system.

There was some talk earlier about an improved office suite made by a partner? We could even utilize/support this ourselves in new document creation. To be following the same line as _normal_ Linux. But if we wish to deviate again...

pvuorela: In the unlikely event that there would be seen big enough benefit, the question can be always revisited. At the moment we don't even have much support for creating new files (well, happens if you take a picture or download a file, but that's implicit).

For app usage, as also mentioned I can use the template from the file manager, but for doing new file inside the app at least LibreOffice doesn't pick up files from templates at all and uses its own template system. I think that's all for the better and would suggest similar thing for other apps. IMO the whole template directory thing is pretty useless.

LaakkonenJussi: Lets deviate then. Changed the PR.

permissions/Base.permission
spiiroin: NB That "Desktop in WebView.permission" is there due to: initial rules were generated semi-automated and in a mass -> it eliminated an access failure from logs. If it is known to be wrong / lack of it would be inconsequential, it should be removed.

README.md
```diff
@@ -115,8 +115,10 @@ Permissions that applications may use (names are subject to change):
 | Music | Access to Music directory, playlists and coverart cache. |
 | Phone | Make Phone calls, either directly or through system voice call UI. |
 | Pictures | Access to Pictures directory and thumbnails. |
+| PublicDir | Access to Public directory. |
 | Sharing | Use of sharing pages in the application to share data via Bluetooth, email or other accounts. |
 | Synchronization | Access to synchronization framework. |
+| UserDirs | Access to Documents, Downloads, Music, Pictures and Public directories. |
```
Tomin1, line 121: Videos too.

Tomin1 (approved):
Still LGTM

permissions/Base.permission
Thaodan: > Unless we want to give access via that to the `$HOME/.local/share/applications` to add URL desktop entries there to be visible on _our_ desktop?

I don't think that would make sense since these files should be in the paths which the xdg-menu spec specifies:
https://specifications.freedesktop.org/menu-spec/menu-spec-latest.html#paths

Also the intend behind this is different IMHO. It is more a place to put desktop files there that are already in the menu or link common folder etc.

If we want to support adding new desktop entries that would be another task see:
https://specifications.freedesktop.org/menu-spec/menu-spec-latest.html#menu-editing

LaakkonenJussi:
@okodron Still waiting for approval from OMP side.

okodron (approved):

PR #63, opened 2021-04-13
Adjust qmf socket path.

Tomin1:
Messageserver (qmf) socket was moved to RUNUSER directory and rules need
adapting to that. Blacklist it by default so that applications without
access to email can't use it.

### DIFF
--- a/permissions/Email.permission
+++ b/permissions/Email.permission
@@ -19,7 +19,8 @@ mkdir     ${HOME}/Downloads/mail_attachments
 whitelist ${HOME}/Downloads/mail_attachments
 
 # MessageServer socket
-whitelist /tmp/qcop-server-0
+noblacklist ${RUNUSER}/messageserver
+read-only   ${RUNUSER}/messageserver
 
 # Messaging framework
 mkdir     ${HOME}/.qmf
--- a/permissions/Base.permission
+++ b/permissions/Base.permission
@@ -30,6 +30,10 @@ include disable-passwdmgr.inc
 # include disable-programs.inc
 # include disable-xdg.inc
 
+# Runuser blacklisting
+mkdir     ${RUNUSER}/messageserver
+blacklist ${RUNUSER}/messageserver
+
 ### home directory whitelisting
 whitelist ${HOME}/.cursors/default
 whitelist ${HOME}/.local/share/glib-2.0/schemas


permissions/Email.permission
```diff
@@ -19,7 +19,9 @@ mkdir     ${HOME}/Downloads/mail_attachments
 whitelist ${HOME}/Downloads/mail_attachments
 
 # MessageServer socket
-whitelist /tmp/qcop-server-0
+mkdir       ${RUNUSER}/messageserver
```
spiiroin, line 22: Have you tested whether it works without doubling the mkdir?
I tend to think that it should. IIRC each mkdir is executed twice: first during parsing and again during sandbox setup along with white/blaclist etc -> even if Base is read last, 1st pass of mkdir in it ought to have chance to do its thing before rules from Email are acted upon.

spiiroin (approved):
LGTM as-is. That possible 2xmkdir thing is probably insignificant even if it could be avoided.

permissions/Email.permission
Tomin1: The thing I'm concerned is that if the blacklisted directory did not exists before the sandbox was constructed, the directory will become accessible when it's created. I tested this now and that seems to be the case. Since messageserver5 is only started on demand and the first sandbox constructed is voicecall-ui, then that means in practice it's always voicecall-ui sandboxing that's also creating it. I guess the alternative might be that we'd construct the directory when user session is started and then we wouldn't need any mkdir rules.

spiiroin: Trying to clarify: IIRC each "mkdir" and "mkfile" op is acted on twice. First during command line arg / config file processing - when firejail is still acting as more or less regular user process in global namespace - and later on also from within sandboxing context -> even if a "mkdir" option is the last thing given to firejail, an attempt is made to create that directory regardless of other rules. And having the same "mkdir"  twice in config files means firejail tries to create that directory four times -> maybe it does not accumulate to much, but it is still excess when/if having it in one place (Base.permission?) would suffice.

Tomin1: Ah, yes. Now I see what you mean. I did some quick testing and yes, having mkdir only in Base.permission suffices.

Tomin1:
@okodron would you like to review?

permissions/Email.permission
```diff
@@ -19,7 +19,8 @@ mkdir     ${HOME}/Downloads/mail_attachments
 whitelist ${HOME}/Downloads/mail_attachments
 
 # MessageServer socket
-whitelist /tmp/qcop-server-0
+noblacklist ${RUNUSER}/messageserver
```
okodron, line 22: Can you noblacklist just a particular `${RUNUSER}/messageserver/qcop-server-0` but not the whole directory?

Tomin1: I think I can't. blacklist rule prevents me from accessing that directory, noblacklist will prevent that rule, so if I say
```
blacklist ${RUNUSER}/messageserver
noblacklist ${RUNUSER}/messageserver/qcop-server-0
```
Then I don't get access to that messageserver directory still which means the files inside are not accessible either.

This directory was added for the socket so this noblacklist rule doesn't allow access to anything else than the socket.

okodron: Ok, just has worries that something evil can deliberately be landed in this directory.

okodron (approved):

permissions/Email.permission
Tomin1: We need to anyway restrict further this ${RUNUSER} access (change from blacklist rules to whitelist rules), there is separate task for that.

PR #64, opened 2021-04-15
Remove application specific paths from Email.permission

Tomin1:
Remove application specific paths from Email.permission. These are
either legacy and not used anymore, were never used, or they are
automatically allowed by sailjail.

Those mail_attachments rules didn't do anything as the directory access
is allowed anyway if app has Downloads permission. Also added a comment
about that dbus rule.


Tomin1:
@okodron

dseight:
LGTM

okodron (approved):

rainemak (approved):
Approving.

PR #66, opened 2021-04-15
Split UDisk2 permissions, Use disable-mnt to control data storage perms.

LaakkonenJussi:
[permissions] Use disable-mnt to control data storage perms, move udisk2 include. OMP#JOLLA-49

Block access to any media mount with disable-mnt in Base.permissions.
Grant permissions to all media when DataStorages permission is used,
including UDisks2 D-Bus.

Move the UDisks permission include away from Base.permissions. This is
to be included separately or via DataStorages permission.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,7 @@ Permissions that applications may use (names are subject to change):
 | Phone | Make Phone calls, either directly or through system voice call UI. |
 | Pictures | Access to Pictures directory and thumbnails. |
 | PublicDir | Access to Public directory. |
+| RemovableMedia | Use memory cards and USB sticks. |
 | Sharing | Use of sharing pages in the application to share data via Bluetooth, email or other accounts. |
 | Synchronization | Access to synchronization framework. |
 | UserDirs | Access to Documents, Downloads, Music, Pictures, Public and Video directories. |
@@ -127,7 +128,6 @@ Internal permissions that applications generally should not use directly:
 | Base       | Base set of permissions that every application is granted implicitly. |
 | CaptivePortal |
 | Connman |
-| DataStorages |
 | GnuPG |
 | FingerprintSensor |
 | Microphone |
--- a/README.md
+++ b/README.md
@@ -135,7 +135,8 @@ Internal permissions that applications generally should not use directly:
 | PinQuery |
 | Sensors |
 | Thumbnails |
-| UDisks |
+| UDisks | Permissions to call UDisks functions, includes UDisksListen. |
+| UDisksListen | Permissions to listen signals and property changes on UDisks2 interfaces. |
 
 ### Permission metadata format
 
--- a/permissions/DataStorages.permission
+++ b/permissions/DataStorages.permission
@@ -1,7 +1,0 @@
-# -*- mode: sh -*-
-
-# x-sailjail-translation-catalog =
-# x-sailjail-translation-key-description =
-# x-sailjail-description = Removable media
-# x-sailjail-translation-key-long-description =
-# x-sailjail-long-description = Use memory cards and USB sticks
--- a/permissions/RemovableMedia.permission
+++ b/permissions/RemovableMedia.permission
@@ -1,0 +1,12 @@
+# -*- mode: sh -*-
+
+# x-sailjail-translation-catalog = sailjail-permissions
+# x-sailjail-translation-key-description = permission-la-removable_media
+# x-sailjail-description = Removable media
+# x-sailjail-translation-key-long-description = permission-la-removable_media_description
+# x-sailjail-long-description = Use memory cards and USB sticks
+
+### Allow to listen/talk to udisks2
+include /etc/sailjail/permissions/UDisksListen.permission
+
+ignore disable-mnt
--- a/permissions/UDisks.permission
+++ b/permissions/UDisks.permission
@@ -1,16 +1,19 @@
 # -*- mode: sh -*-
 
-# FIXME: what exactly requires udisks???
-
 # x-sailjail-translation-catalog =
-# x-sailjail-translation-key-description =
-# x-sailjail-description = Use UDisks2
-# x-sailjail-translation-key-long-description =
-# x-sailjail-long-description =
+# x-sailjail-translation-key-description = permission-la-udisks2
+# x-sailjail-description = Manage and access storage devices
+# x-sailjail-translation-key-long-description = permission-la-udisks2_description
+# x-sailjail-long-description = Permission to access the storage devices, allows to call device D-Bus methods.
 
-# BEG systembus-org.freedesktop.UDisks2.resource
-dbus-system.talk org.freedesktop.UDisks2
-dbus-system.broadcast org.freedesktop.UDisks2=org.freedesktop.UDisks2.*@/*
-dbus-system.talk org.freedesktop.UDisks2.*
-dbus-system.broadcast org.freedesktop.UDisks2.*=org.freedesktop.UDisks2.*@/*
-# END systembus-org.freedesktop.UDisks2.resource
+include /etc/sailjail/permissions/UDisksListen.permission
+
+# BEG systembus-org.freedesktop.UDisks2 method call access
+dbus-system.call org.freedesktop.UDisks2=org.freedesktop.DBus.Properties.Set@/*
+dbus-system.call org.freedesktop.UDisks2=org.freedesktop.UDisks2.Block.Format@/*
+dbus-system.call org.freedesktop.UDisks2=org.freedesktop.UDisks2.Block.Rescan@/*
+dbus-system.call org.freedesktop.UDisks2=org.freedesktop.UDisks2.Encrypted.Lock@/*
+dbus-system.call org.freedesktop.UDisks2=org.freedesktop.UDisks2.Encrypted.Unlock@/*
+dbus-system.call org.freedesktop.UDisks2=org.freedesktop.UDisks2.FileSystem.Mount@/*
+dbus-system.call org.freedesktop.UDisks2=org.freedesktop.UDisks2.FileSystem.Unmount@/*
+# END systembus-org.freedesktop.UDisks2 method call access
--- a/permissions/UDisksListen.permission
+++ b/permissions/UDisksListen.permission
@@ -1,0 +1,22 @@
+# -*- mode: sh -*-
+
+# x-sailjail-translation-catalog =
+# x-sailjail-translation-key-description = permission-la-udisks2_listen
+# x-sailjail-description = Access storage device information
+# x-sailjail-translation-key-long-description = permission-la-udisks2_description
+# x-sailjail-long-description = Permissions to access properties and to listen to property changes on storage devices
+
+# BEG systembus-org.freedesktop.UDisks2.resource
+dbus-system.broadcast org.freedesktop.UDisks2=org.freedesktop.UDisks2.*@/*
+dbus-system.broadcast org.freedesktop.UDisks2.*=org.freedesktop.UDisks2.*@/*
+
+# END systembus-org.freedesktop.UDisks2.resource
+# BEG systembus-org.freedesktop.UDisks2 call access
+### Allow to get available functions and access properties with getters
+dbus-system.call org.freedesktop.UDisks2=org.freedesktop.DBus.Introspectable.Introspect@/*
+dbus-system.call org.freedesktop.UDisks2=org.freedesktop.DBus.Properties.Get@/*
+dbus-system.call org.freedesktop.UDisks2=org.freedesktop.DBus.Properties.GetAll@/*
+### Not listed in udisks2defines.h but exists in the nemo-qml-plugin-systemsettings code and is thus required by many apps
+### TODO: remove when nemo-qml-plugin-systemsettings do not need this anymore
+dbus-system.call org.freedesktop.UDisks2=org.freedesktop.UDisks2.Manager.GetBlockDevices@/org/freedesktop/UDisks2/Manager
+# END systembus-org.freedesktop.UDisks2 call access
--- a/permissions/Base.permission
+++ b/permissions/Base.permission
@@ -158,8 +158,8 @@ seccomp
 ### notifications
 include /etc/sailjail/permissions/Notifications.permission
 
-### udisks2
-include /etc/sailjail/permissions/UDisks.permission
+### Disable access to media by default
+disable-mnt
 
 ### privileged data
 # make sure privileged data directory exists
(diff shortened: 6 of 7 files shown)


rainemak:
Nice! Need to check in more detail but that goes tomorrow.

Please combine references to one single line:
```
[permissions] Use disable-mnt to control data storage perms, move udisks2 include. JB#53818 OMP#JOLLA-49
```

permissions/DataStorages.permission
```diff
@@ -5,3 +5,9 @@
 # x-sailjail-description = Removable media
 # x-sailjail-translation-key-long-description =
```
Tomin1, line 6: Please fill in these values (catalog, key-description, key-long-description).

permissions/DataStorages.permission
```diff
@@ -5,3 +5,9 @@
...
 # x-sailjail-long-description = Use memory cards and USB sticks
+
+### Allow to listen/talk to udisks2
+# Used for mounting, formatting
```
Tomin1, line 10: I wonder if we should allow this at all. Then again, we need to check that UDisks.permission still anyway, so it's not too critical to do it now.

Tomin1 (changes requested):
Also please fill in description of DataStorages to README. I think this could be also moved from internal list to "public" list.

BTW, if you can think of a better name than DataStorages, that could be also changed now.

permissions/DataStorages.permission
LaakkonenJussi: Ah, since this file was already there I just missed that part, will add.

permissions/DataStorages.permission
LaakkonenJussi: That was what I was thinking most of the day that should some app have the permission to add/format sd/usb or is it completely internal functionality.

LaakkonenJussi:
> Also please fill in description of DataStorages to README. I think this could be also moved from internal list to "public" list.
> 
> BTW, if you can think of a better name than DataStorages, that could be also changed now.

Yeah well functionality first, documentation second. Early review gets the bugs etc. On it!

permissions/DataStorages.permission
```diff
@@ -5,3 +5,9 @@
...
+include /etc/sailjail/permissions/UDisks.permission
+
+ignore disable-mnt
```
rainemak, line 13: Wondering what else gets exposed with this other than external storages?

LaakkonenJussi: No need to wonder, [the manual pages ](https://firejail.wordpress.com/features-3/man-firejail-profile/) just state this clearly:
`
       disable-mnt
              Disable /mnt, /media, /run/mount and /run/media access.
`

Unless the curtains do not match the drapes?

permissions/DataStorages.permission
spiiroin: Getting it off "Base" would already be a massive improvement.

But there was some reason why it got included in there... I do not remember exactly what it might have been, but typical reason for illogically lavish Base permission would be: without it some mw component crashes on unchecked access failure -> changes like this need testing / enough time in devel.

LaakkonenJussi: But could we limit the scope of `UDisks2` here as well? There are some a bit dangerous D-Bus methods that can be used with this lavish rule, Delete(), Resize() do not sound like our mw apps do really need. Settings is the only one that can manage removable media and it does not yet use sailjail/firejail?

spiiroin: Well, making it stricter would of course be better. With assumption like "only settings modifies" it could even be only  enumeration/getter functionality allowed or stte.

LaakkonenJussi: I went through https://git.sailfishos.org/mer-core/nemo-qml-plugin-systemsettings/blob/master/src/udisks2defines.h and added the more specific interface/method rules with 3d6d780 + the `org.freedesktop.UDisks2.Manager.GetBlockDevices` which is in the code only and what has been complained in the journal if there are no UDisks2 permissions.

permissions/UDisks.permission
```diff
@@ -1,16 +1,18 @@
...
-
-# x-sailjail-translation-catalog =
-# x-sailjail-translation-key-description =
+# x-sailjail-translation-catalog = sailfish-catalog
+# x-sailjail-translation-key-description = permission-la-udisks2
 # x-sailjail-description = Use UDisks2
-# x-sailjail-translation-key-long-description =
-# x-sailjail-long-description =
+# x-sailjail-translation-key-long-description = permission-la-udisks2_description
+# x-sailjail-long-description = Access signals and properties on UDisks2, allows to get the devices, lock, mount and format to be called.
```
Tomin1, line 7: I would not mention UDisks2 here. This is user visible text (when included directly by an application) and they don't have any idea what that means.

permissions/UDisksListen.permission
```diff
@@ -0,0 +1,18 @@
+# -*- mode: sh -*-
+
+# x-sailjail-translation-catalog = sailfish-catalog
+# x-sailjail-translation-key-description = permission-la-udisks2_listen
+# x-sailjail-description = Use UDisks2 for listening
```
Tomin1, line 5: Similarly here: there shouldn't be UDisks2 here. Perhaps this shouldn't have any translations and instead it should be an internal permission that's just included from RemovableMedia.

Tomin1 (commented):
I left some comments to think about.

README.md
```diff
@@ -115,6 +115,7 @@ Permissions that applications may use (names are subject to change):
 | Music | Access to Music directory, playlists and coverart cache. |
 | Phone | Make Phone calls, either directly or through system voice call UI. |
 | Pictures | Access to Pictures directory and thumbnails. |
+| RemovableMedia | Use memory cards and USB sticks |
```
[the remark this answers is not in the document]
LaakkonenJussi: Added `.` with 94588c1489e297c32c6da6ee193a4119ed181cc6

permissions/UDisks.permission
LaakkonenJussi: Yeah, that might be better now with 94588c1489e297c32c6da6ee193a4119ed181cc6 ?

permissions/UDisksListen.permission
LaakkonenJussi: Better with 94588c1489e297c32c6da6ee193a4119ed181cc6 ?

LaakkonenJussi:
> The format that needs to be used is:
> 
> ```
> dbus.system.[call|broadcast] <NAME>[.*]=<INTERFACE>[.*][@<OBJECT>[/*]]
> ```
> 
> There is no '=' in any of these call rules -> invalid rules for never-going-to-exist names / stte.
> 
> The documentation / even code is not always very clear, but you can do at least format checking with e.g.:
> 
> ```
> $ xdg-dbus-proxy 3>/dev/null --fd=3 fooname barsocket --call=org.freedesktop.UDisks2.Block.Format
> 'org.freedesktop.UDisks2.Block.Format' is not a valid name + rule
> ```
> 
> = not ok
> 
> ```
> $ xdg-dbus-proxy 3>/dev/null --fd=3 fooname barsocket --call=org.freedesktop.UDisks2.Block=org.freedesktop.UDisks2.Block.Format
> ^C
> ```
> 
> = ok

Thanks for clarifying my head with this as the documentation did not. Fixed these issues in the files.

permissions/DataStorages.permission
LaakkonenJussi: I did split the permissions to `UDisks` and to `UDisksListen` - the latter of which is included by `RemovableMedia` and by `UDisks`. This way the apps can simply get info on the removable media, as well as other storage devices on the system but not to tamper with them at any level. All destructive, mounting, etc. are left in `UDisks` permission.

permissions/DataStorages.permission
LaakkonenJussi: Done, so resolving.

permissions/UDisks.permission
```diff
@@ -1,16 +1,18 @@
 # -*- mode: sh -*-
 
-# FIXME: what exactly requires udisks???
+# x-sailjail-translation-catalog = sailfish-catalog
+# x-sailjail-translation-key-description = permission-la-udisks2
+# x-sailjail-description = Manage storage devices
+# x-sailjail-translation-key-long-description = permission-la-udisks2_description
+# x-sailjail-long-description = Permission to access signals and properties on storage devices, allows to call devices get, lock, mount and format functionality.
```
Tomin1, line 7: I wouldn't talk about signals or properties here either because those are do not mean anything for user. These descriptions are something user sees. We can describe these separately for developer in README or here as comments.

While this is internal and will not be available in application store, this has translation catalog which means that these will generate translatable strings and translators will translate them. So we should consider that too if we add any texts here. It's also possible to leave that translation catalog empty and it will just fallback to the values here.

LaakkonenJussi:
Updated the PR and squashed the changes to two separate commits, one for UDisks2 and one for the RemovableMedia. @Tomin1 @spiiroin @okodron I guess this is ready for final review. Checked with sailjail --trace that D-Bus log did not anymore contain any `DENIED` lines when trying to access removable media.

permissions/UDisksListen.permission
```diff
@@ -0,0 +1,23 @@
+# -*- mode: sh -*-
+
+# x-sailjail-translation-catalog = sailfish-catalog
+# x-sailjail-translation-key-description = permission-la-udisks2_listen
+# x-sailjail-description = Access storage device information
+# x-sailjail-translation-key-long-description = permission-la-udisks2_description
+# x-sailjail-long-description = Permissions to listen signals and property changes on storage devices
```
Tomin1, line 7: What I said applies also here, and even more so if this is in the public list, although I would probably put this to internal list and make it available only together with RemovableMedia.

permissions/UDisks.permission
LaakkonenJussi: Right. Lets just keep that empty and mention this only in the internals section.

permissions/UDisksListen.permission
LaakkonenJussi: Better to have this only with `RemovableMedia` permission, will change.

permissions/UDisks.permission
LaakkonenJussi: 6fbe565b136b84ccbc77dc5d833ed6cdea3663fa ok?

permissions/UDisksListen.permission
LaakkonenJussi: 6fbe565b136b84ccbc77dc5d833ed6cdea3663fa ok?

Tomin1: Almost, please drop those translation-catalog values still so that this won't generate translatable strings. I don't think we should have those for internal permissions unless they are intended to be used separately.

LaakkonenJussi: I was about to ask some point that what is the functionality of that but it makes complete sense now. Removed.

Tomin1: Erm, sorry. That translation-catalog key doesn't affect that what I thought it was when I wrote that comment. The idea behind that key is that we can specify in permission file the translation catalog that middleware should read. However currently the only place that defines any permission files is this repository so in practice it's always the same value. Now I realized that translation generator script doesn't use that value at all (although it could check that the value is "sailjail-permissions" for correctness). So it's probably best to not to define any strings here. Alternatively we could improve the script a little.

LaakkonenJussi: Dropped with bec9f7dd73014d68e81c6aa6c7f95119fe8fac45

LaakkonenJussi: After a tonload of confusion and adding of 95a38b66ea3843e83dc300b4331b16a4b0ae5668 to the commits I guess it is now as it should or was intended to be.

Tomin1: I think it's great. :+1:

permissions/RemovableMedia.permission
```diff
@@ -0,0 +1,13 @@
+# -*- mode: sh -*-
+
+# x-sailjail-translation-catalog = sailjail-permissions
+# x-sailjail-translation-key-description = permission-la-removable_media
+# x-sailjail-description = Removable media
+# x-sailjail-translation-key-long-description = permission-la-removable_media_description
+# x-sailjail-long-description = Use memory cards and USB sticks
+
+### Allow to listen/talk to udisks2
+# Used for mounting, formatting, locking only
```
Tomin1, line 10: Perhaps this comment should be removed still.

Tomin1 (approved):
I think this is fine.

permissions/RemovableMedia.permission
LaakkonenJussi: Oh yeah, that is a remnant that should be removed.

LaakkonenJussi: Removed by editing the relevant commit and force pushing. Resolving.

LaakkonenJussi:
@okodron waiting for review on your side as well.

okodron (approved):

permissions/UDisks.permission
```diff
@@ -1,16 +1,19 @@
...
-# BEG systembus-org.freedesktop.UDisks2.resource
-dbus-system.talk org.freedesktop.UDisks2
-dbus-system.broadcast org.freedesktop.UDisks2=org.freedesktop.UDisks2.*@/*
-dbus-system.talk org.freedesktop.UDisks2.*
-dbus-system.broadcast org.freedesktop.UDisks2.*=org.freedesktop.UDisks2.*@/*
-# END systembus-org.freedesktop.UDisks2.resource
+# BEG systembus-org.freedesktop.UDisks2 method call access
+dbus-system.call org.freedesktop.UDisks2=org.freedesktop.DBus.Properties.Set@/*
+dbus-system.call org.freedesktop.UDisks2=org.freedesktop.UDisks2.Block.Format@/org/freedesktop/UDisks2/Block
+dbus-system.call org.freedesktop.UDisks2=org.freedesktop.UDisks2.Block.Rescan@/org/freedesktop/UDisks2/Block
```
spiiroin, line 14: Are these object paths right?

... I have no idea, but with "qdbus --system org.freedesktop.UDisks2" I get objects like "/org/freedesktop/UDisks2/block_devices/mmcXYZ" that then do have Block interface available.

LaakkonenJussi: Ah yes, maybe the interfaces could be dropped here as well for the method calls.

LaakkonenJussi: Because, there are many different paths that make these same calls available.

LaakkonenJussi: Like e751dcf3128d721bc18314458dcd64e7059e6c4c

spiiroin (approved):

PR #67, opened 2021-04-19
Add CommunicationHistory.permission.

Tomin1:
Remove CommHistory specific permissions from Phone and Messages and move
those to separate CommunicationHistory.permission file. Allow only relevant D-Bus
interfaces.

Remove those old resource comments and add more useful comments for the
purposes of the rules.

Move settings permission outside Phone and Messages permission as
it doesn't belong here as it's about being able to open a Settings
page which is an application specific thing to do.

### DIFF
--- a/permissions/Phone.permission
+++ b/permissions/Phone.permission
@@ -6,40 +6,23 @@
 # x-sailjail-translation-key-long-description = permission-la-phone_description
 # x-sailjail-long-description = Make phone calls
 
-# BEG systembus-org.ofono.resource
+# Ofono to place calls etc.
 dbus-system.talk org.ofono
 dbus-system.broadcast org.ofono=org.ofono.*@/*
 dbus-system.broadcast org.ofono=org.nemomobile.ofono.*@/*
-# END systembus-org.ofono.resource
 
 # Voicecall service
 dbus-user.talk org.nemomobile.voicecall
 dbus-user.broadcast org.nemomobile.voicecall=org.nemomobile.voicecall.*@/*
-
-# BEG sessionbus-org.freedesktop.Telepathy.resource
 dbus-user.talk org.freedesktop.Telepathy
 dbus-user.broadcast org.freedesktop.Telepathy=org.freedesktop.Telepathy.*@/*
 dbus-user.talk org.freedesktop.Telepathy.*
 dbus-user.broadcast org.freedesktop.Telepathy.*=org.freedesktop.Telepathy.*@/*
-# END sessionbus-org.freedesktop.Telepathy.resource
 
-mkdir     ${HOME}/.local/share/commhistory
-whitelist ${HOME}/.local/share/commhistory
-
-# BEG sessionbus-org.nemomobile.CommHistory.resource
-dbus-user.talk org.nemomobile.CommHistory
-dbus-user.broadcast org.nemomobile.CommHistory=org.nemomobile.CommHistory.*@/*
-# END sessionbus-org.nemomobile.CommHistory.resource
-
-# BEG systembus-org.ofono.SmartMessagingAgent.resource
-dbus-system.talk org.ofono.SmartMessagingAgent
-# END systembus-org.ofono.SmartMessagingAgent.resource
-
-# BEG sessionbus-com.jolla.voicecall.ui.resource
+# Open Phone UI to make calls
 dbus-user.talk com.jolla.voicecall.ui
-# END sessionbus-com.jolla.voicecall.ui.resource
 
-# Have audio permissions implicitly.
+# Have audio permissions implicitly
 include /etc/sailjail/permissions/Audio.permission
 
 # Allow requesting sim pin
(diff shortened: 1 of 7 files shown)


permissions/CallLogs.permission
```diff
@@ -0,0 +1,14 @@
...
+# x-sailjail-description = Call history
+# x-sailjail-translation-key-long-description = permission-la-calllogs_description
+# x-sailjail-long-description = View and edit previous calls and messages
+
+mkdir     ${HOME}/.local/share/commhistory
+whitelist ${HOME}/.local/share/commhistory
+
+dbus-user.talk org.nemomobile.CommHistory
+dbus-user.broadcast org.nemomobile.CommHistory=org.nemomobile.CommHistory.*@/org/nemomobile/CommHistory
+dbus-user.call org.nemomobile.CommHistory=org.nemomobile.CommHistory.*@/org/nemomobile/CommHistory
```
Tomin1, line 14: I wonder if any of the other interfaces in org.nemomobile.CommHistory would be useful here. They seem a little disconnected to me and at least for pure call and message history point of view missing those interfaces didn't seem to affect anything.

```
$ /usr/lib/qt5/bin/qdbus org.nemomobile.CommHistory
/
/CommHistoryModel
/org
/org/freedesktop
/org/freedesktop/Telepathy
/org/freedesktop/Telepathy/Client
/org/freedesktop/Telepathy/Client/CommHistory
/org/nemomobile
/org/nemomobile/AccountPresence
/org/nemomobile/CommHistory
```

Those telepathy interfaces actually belong to another service name (but they are exposed through this due to limitations of QDBus) but those /CommHistoryModel and /org/nemomobile/CommHistory make me wonder.

permissions/Phone.permission
```diff
@@ -6,46 +6,28 @@
...
-mkdir     ${HOME}/.local/share/commhistory
-whitelist ${HOME}/.local/share/commhistory
 
-# BEG sessionbus-org.nemomobile.CommHistory.resource
-dbus-user.talk org.nemomobile.CommHistory
-dbus-user.broadcast org.nemomobile.CommHistory=org.nemomobile.CommHistory.*@/*
-# END sessionbus-org.nemomobile.CommHistory.resource
-
-# BEG systembus-org.ofono.SmartMessagingAgent.resource
+# ???
```
Tomin1, line 22: Why is SmartMessagingAgent here?

permissions/CallLogs.permission
```diff
@@ -0,0 +1,14 @@
+# -*- mode: sh -*-
+
+# x-sailjail-translation-catalog = sailjail-permissions
+# x-sailjail-translation-key-description = permission-la-calllogs
+# x-sailjail-description = Call history
```
Tomin1, line 5: While this is also about message history, I used word call here. I wonder if there was some nice way of combining them. "Call and message history" is a bit long IMO and translations can be even longer.

Also that could affect permission file naming if we come up with something better here.

spiiroin: Well, commhistory = "Communications event history" -> maybe there would be something not too clumsy / abstract in that direction? But I'm not sure if  "CommunicationHistory" fits that bill.

permissions/Phone.permission
spiiroin: The process was something like: On dbus errors/complaints about a service -> service.resource file was created. These were then included in application profiles until errors went away. When all apps were gotten to working state, the resulting resource files were bundled up as feature.permission files, while trying to keep number of "features" relatively low -> done en masse -> feel free to assume that there are logical errors.

permissions/CallLogs.permission
spiiroin: There are couple of these "one processes, multiple services" cases that were really difficult to understand from bottom up, errors to introspection point of view as one would need to dig deep into source code in order to grok how it ought to be (which is what should drive these rules) -> what there is about such services is more or less guesswork.

spiiroin (approved):
Looks ok, did not spot any obvious faults. I'll leave the file name up to you.

permissions/Phone.permission
Tomin1: Ok, well... I removed that and let's see what happens. libofono seems to suggest more contacts related reason for having this.

permissions/CallLogs.permission
Tomin1: Yes, that's even longer, at least in English. Well, I didn't change this quite yet to anything else. And perhaps that call history is still understandable enough with that longer explanation.

permissions/CallLogs.permission
Tomin1: Looks like /org/nemomobile/AccountPresence is not in libcommhistory at least but [/CommHistoryModel path is](https://git.sailfishos.org/mer-core/libcommhistory/blob/master/src/constants.h#L31) although it's not really clear if it could be useful on application side ([the only place where it uses the string](https://git.sailfishos.org/mer-core/libcommhistory/blob/master/src/updatesemitter.cpp)).

permissions/CallLogs.permission
Tomin1: It seems there should be quite enough space for "communication history" so I'll use that for now. Changed that file name too.

Tomin1:
@okodron or @dseight, would either of you check this? Thanks.

okodron (approved):

Tomin1:
Rebased.

dseight (approved):

PR #69, opened 2021-04-21
[permissions] Add sailfish-share to Sharing permissions. JB#53987 OMP#JOLLA-75

blammit:


permissions/Sharing.permission
```diff
@@ -32,6 +32,11 @@ dbus-user.broadcast com.jolla.PinQuery=com.jolla.PinQuery.*@/*
 dbus-user.talk com.jolla.Connectiond
 dbus-user.broadcast com.jolla.Connectiond=com.jolla.Connectiond.*@/*
 
+# Sharing process
+whitelist /usr/share/sailfish-share
```
spiiroin, line 36: Could you elaborate on what is happening in this directory?

pvuorela: For the sharing dialog UI don't think it belongs here. First as "app specific" thing it could be more of a .profile, but then again as system UI not sure do we even want to sandbox it.

blammit: Ah yes, this is longer necessary as sandboxing was removed elsewhere. Updated to remove this now.

spiiroin (approved):
Now that the puzzling whitelist is gone, I can't see any faults in this.

Tomin1 (approved):

Tomin1:
@okodron or @dseight, please review.

okodron (approved):

dseight (approved):

PR #70, opened 2021-04-23
Move pin query to its own permission

Tomin1:
Move pin query dialog opening to its own permission but include that
from Phone and Messages. Also restrict this to pin query only as that
service name exposes some other interfaces as well.

Add a comment that this D-Bus API requires privileged access to
function. This permission is internal without any descriptions. As this
is only used through Phone and Messages those are not needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,7 @@ Internal permissions that applications generally should not use directly:
 | FingerprintSensor |
 | Microphone |
 | Notifications |
+| PinQuery |
 | Sensors |
 | Thumbnails |
 | UDisks |
--- a/permissions/PinQuery.permission
+++ b/permissions/PinQuery.permission
@@ -1,0 +1,5 @@
+# Privileged D-Bus API
+# Allows to open pin query dialog
+dbus-user.talk com.jolla.PinQuery
+dbus-user.call com.jolla.PinQuery=com.jolla.PinQuery.requestSimPin@/com/jolla/PinQuery
+dbus-user.broadcast com.jolla.PinQuery=com.jolla.PinQuery.requestCanceled@/com/jolla/PinQuery
--- a/permissions/Messages.permission
+++ b/permissions/Messages.permission
@@ -35,5 +35,4 @@ dbus-system.talk org.ofono.SmartMessagingAgent
 # END systembus-org.ofono.SmartMessagingAgent.resource
 
 # Allow requesting sim pin
-dbus-user.talk com.jolla.PinQuery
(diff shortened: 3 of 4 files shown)


Tomin1:
@okodron or @dseight, please review.

permissions/PinQuery.permission
```diff
@@ -0,0 +1,13 @@
+# -*- mode: sh -*-
+
+# x-sailjail-translation-catalog =
+# x-sailjail-translation-key-description =
+# x-sailjail-description =
+# x-sailjail-translation-key-long-description =
+# x-sailjail-long-description =
```
LaakkonenJussi, line 7: Since we're about to soon have the `x-sailjail-translation-catalog` to trigger the use of the translations (#66) shouldn't we put the keys and descriptions now when creating new permission files?

Tomin1: I would argue that if the permission will not be needed separately and it doesn't have already defined texts, we shouldn't even need these lines.

LaakkonenJussi: Ok. Do we need these translation content in each of the files or could these just be omitted from ones that are not going to be needed separately?

Tomin1: We could omit them.

spiiroin (approved):
Logic wise LGTM.

README.md
```diff
@@ -130,6 +130,7 @@ Internal permissions that applications generally should not use directly:
 | FingerprintSensor |
 | Microphone |
 | Notifications |
+| PinQuery |
```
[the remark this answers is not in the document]
Tomin1: Do you mean the term pin query? This is an internal permission that no application should use thus we can rename it any day if there is some collision. Then again, I'm not sure there will be anything else that should use word pin. I think we have tried to keep it so that pin is always referring to SIM pin and anything else is code or something like that.

permissions/PinQuery.permission
Tomin1: Removed the extra cruft from this file. I think it's probably best to do the rest of these at once in a future task. Right now it would just create collisions with other PRs.

README.md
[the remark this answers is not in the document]
spiiroin: Yes. PIN is "personal identification number" which is rather generic term and used also outside sim pin context. So if it were shown to user it might be confusing. But as it is used on internally without it being visible to users -> no need to change / anything.

okodron (approved):

PR #71, opened 2021-04-23
Add Tracker3 miner service to MediaIndexing

pvuorela:
Tracker1 still kept. That can be removed once the update is merged.

For this one shouldn't hurt integrating already before the update.

Related to https://git.sailfishos.org/mer-core/tracker/merge_requests/34

@Tomin1 @rainemak @llewelld

### DIFF
--- a/permissions/MediaIndexing.permission
+++ b/permissions/MediaIndexing.permission
@@ -7,6 +7,7 @@
 # x-sailjail-long-description = List files stored on device
 
 whitelist /usr/share/tracker
+whitelist /usr/share/tracker3
 
 mkdir     ${HOME}/.cache/tracker
 whitelist ${HOME}/.cache/tracker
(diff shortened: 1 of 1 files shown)


permissions/MediaIndexing.permission
```diff
@@ -20,3 +20,6 @@ dbus-user.broadcast org.freedesktop.Tracker1=org.freedesktop.Tracker1.*@/*
 dbus-user.talk org.freedesktop.Tracker1.*
 dbus-user.broadcast org.freedesktop.Tracker1.*=org.freedesktop.Tracker1.*@/*
 # END sessionbus-org.freedesktop.Tracker1.resource
+
+dbus-user.talk org.freedesktop.Tracker3.Miner.Files
```
Tomin1, line 24: We should probably now keep checking if there is some reason to limit access via dbus-user.call. This gives access to everything which can be okay for some services but it's sometimes too lax.

pvuorela: Yea, and actually the whole new tracker is now more aimed for sandboxing. But so far I've aimed more to getting the new version running properly and haven't bothered with the permission side too much. Thinking this shouldn't probably be much of new security problem so could be an intermediate step.

Tomin1 (approved):
LGTM as an intermediate step

okodron (approved):

pvuorela:
Rebased and added also /usr/share/tracker3 which is needed on some cases too.

PR #73, opened 2021-04-30
Restrict audio permissions, update Microphone perms.

LaakkonenJussi:
Add use of `nosound` to block access to any audio device, including Pulseaudio. Remove this with Audio permissions.

Add mere translations for Microphone permissions as there are no means to separate playback and recording as of now. Microphone can be used as a separate permission, which as of now allows also playback of audio since `Audio` permission is included.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ Permissions that applications may use (names are subject to change):
 | Ambience   | Set and edit ambiences. |
 | AppLaunch  | Launching and stopping systemd services. This is usually needed for background tasks. |
 | ApplicationInstallation | Installing and uninstalling applications. |
-| Audio | Playing audio, changing audio configuration and showing audio controls on lockscreen. |
+| Audio | Playing and recording audio (since Pulseaudio streams cannot be separated both are enabled with this, but it is subject to change), changing audio configuration and showing audio controls on lockscreen. |
 | Bluetooth | Connecting to and using Bluetooth and NFC hardware. |
 | Calendar | Display and editing of calendar events. |
 | CallRecordings | Access recorded calls. |
@@ -112,7 +112,7 @@ Permissions that applications may use (names are subject to change):
 | Location | Use GPS and positioning. |
 | MediaIndexing | Access to Tracker to list files on device. If you have access to a data directory, you may want to use also this. |
 | Messages | Access to message data and to send SMS messages. |
-| WebView | If you use Gecko based WebView you need this. |
+| Microphone | Record audio with microphone. Use Audio permission for playback of the recorded audio (but since Pulseaudio streams cannot be separated this enables also audio playback, which is subject to change). |
 | Music | Access to Music directory, playlists and coverart cache. |
 | Phone | Make Phone calls, either directly or through system voice call UI. |
 | Pictures | Access to Pictures directory and thumbnails. |
@@ -122,6 +122,7 @@ Permissions that applications may use (names are subject to change):
 | Synchronization | Access to synchronization framework. |
 | UserDirs | Access to Documents, Downloads, Music, Pictures, Public and Video directories. |
 | Videos | Access to Videos directory and thumbnails. |
+| WebView | If you use Gecko based WebView you need this. |
 
 Internal permissions that applications generally should not use directly:
 | Permission | Description |
@@ -131,7 +132,6 @@ Internal permissions that applications generally should not use directly:
 | Connman |
 | GnuPG |
 | FingerprintSensor |
-| Microphone |
 | Notifications |
 | PinQuery |
 | Sensors |
--- a/permissions/Audio.permission
+++ b/permissions/Audio.permission
@@ -2,9 +2,16 @@
 
 # x-sailjail-translation-catalog = sailjail-permissions
 # x-sailjail-translation-key-description = permission-la-audio
-# x-sailjail-description = Play audio
+# x-sailjail-description = Play and record audio
 # x-sailjail-translation-key-long-description = permission-la-audio_description
-# x-sailjail-long-description = Play audio and show audio controls on lockscreen
+# x-sailjail-long-description = Play and record audio, and show audio controls on lockscreen
+
+### Do not prevent access to Pulseaudio socket etc. audio devices
+# FIXME: Separete permissions as this gives access to Pulseaudio, Alsa and to microphone.
+# In order to support recording with microphone use the Microphone permission
+# separately as this functionality is subject to change when Pulseaudio streams
+# can be isolated with permissions.
+ignore nosound
 
 private-etc pulse
 
--- a/permissions/Base.permission
+++ b/permissions/Base.permission
@@ -170,6 +170,10 @@ privileged-data -
 whitelist ${HOME}/.config/QtProject/qtlogging.ini
 read-only ${HOME}/.config/QtProject/qtlogging.ini
 
+### Block direct access to sound devices and configs
+# Disables access to pulseaudio by default
+nosound
+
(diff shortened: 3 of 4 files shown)


LaakkonenJussi:
@okodron could you review as well?

permissions/Audio.permission
```diff
@@ -6,6 +6,9 @@
 # x-sailjail-translation-key-long-description = permission-la-audio_description
 # x-sailjail-long-description = Play audio and show audio controls on lockscreen
```
Tomin1, line 7: Since this will give access to microphone, should we mention here that as well? From user perspective that would be nice to know.

For developers we could move that Microphone to the public list and add description there. Also might be useful to describe somewhere (perhaps as a comment in Audio.permission if anyone reads those) that while currently Audio gives access to microphone, developer should require Microphone for compatibility with future releases.

LaakkonenJussi: Generally this gives access to any audio Pulseaudio can handle. Since we have in mind to split the permissions might it be more clearer for the future if we kept the descriptions separate and have only as a comments in the permission files to describe the actual functionality.

However, I see the point that user might want to know that by giving this `Audio` permission to an application it can also access microphone. Perhaps we should change the descriptions when the functionality changes, and also the translations need to be updated then.

Tomin1: Yes, exactly what I meant. Let's change the descriptions again when the functionality changes. And because the intention of the permission is different from what it actually does, we should document that in the comments so that developers can prepare for that change happening later.

LaakkonenJussi: @Tomin1 updated these in 3c6c421 - using "Microphone" in the descriptions might be too long even? Since we've used simply "play audio" I thought that "record audio" might be better in line with that.

"Play audio and record with microphone, and show audio controls ..." seems quite long for a description even.

Tomin1: Yeah, no need to mention that microphone. I think the way you wrote long description is good.

README.md
```diff
@@ -98,7 +98,7 @@ Permissions that applications may use (names are subject to change):
 | Ambience   | Set and edit ambiences. |
 | AppLaunch  | Launching and stopping systemd services. This is usually needed for background tasks. |
 | ApplicationInstallation | Installing and uninstalling applications. |
-| Audio | Playing audio, changing audio configuration and showing audio controls on lockscreen. |
+| Audio | Playing audio, changing audio configuration, recording via microphone and showing audio controls on lockscreen. |
```
Tomin1, line 101: I would perhaps write this here in the way we intend it to work. I.e. no recording at all.

README.md
```diff
@@ -111,7 +111,7 @@ Permissions that applications may use (names are subject to change):
 | Location | Use GPS and positioning. |
 | MediaIndexing | Access to Tracker to list files on device. If you have access to a data directory, you may want to use also this. |
 | Messages | Access to message data and to send SMS messages. |
-| WebView | If you use Gecko based WebView you need this. |
+| Microphone | Record audio with microphone. Gives access to audio devices for playback as well. |
```
Tomin1, line 114: Here I would just mention that "Record audio with microphone." and perhaps recommend Audio for playback.

permissions/Audio.permission
```diff
@@ -2,9 +2,13 @@
...
 # x-sailjail-translation-catalog = sailjail-permissions
 # x-sailjail-translation-key-description = permission-la-audio
-# x-sailjail-description = Play audio
+# x-sailjail-description = Play and record audio
 # x-sailjail-translation-key-long-description = permission-la-audio_description
-# x-sailjail-long-description = Play audio and show audio controls on lockscreen
+# x-sailjail-long-description = Play and record audio, and show audio controls on lockscreen
+
+### Do not prevent access to Pulseaudio socket etc. audio devices
+# This gives access also to microphone but for future compatibility use of Microphone is recommended as well.
```
Tomin1, line 10: Perhaps this could more along the lines of

> While this allows access to microphone, you should require also Microphone permission to ensure access.

If you want, there could "subject to change" or that kind of expressions too.

Tomin1: And of course if you want to explain it further to developer, you could say that pulse audio doesn't have means to control them separately.

README.md
LaakkonenJussi: Well hiding the recording part is kind of misleading in a way, as that is the reality until the streams can be separated one way or another. Maybe just adding that recording in parenthesis with a note "until Pulseaudio streams can be separated" as that is intended to be in the actual permission file. I see no reason to document it here as well as this is what the developer should really read.

permissions/Audio.permission
LaakkonenJussi: Added a lengthy description with a `FIXME` in 36f9ad3 as that separation of Alsa and Pulse permissions seemed to be a viable option in the future.

README.md
LaakkonenJussi: Added with 36f9ad3

README.md
LaakkonenJussi: Removed that in 36f9ad3 but I'm not really happy with it. It may become as a surprise to devs when only using `Audio` permission also mic starts to work.

permissions/Audio.permission
```diff
@@ -2,9 +2,16 @@
 
 # x-sailjail-translation-catalog = sailjail-permissions
 # x-sailjail-translation-key-description = permission-la-audio
-# x-sailjail-description = Play audio
+# x-sailjail-description = Play and record audio
```
spiiroin, line 5: As this is what we show to user: IMO it needs to be in sync with reality -> having "and record" here is a good thing.

But it also means this description is not in sync with what README.md says and I'm wondering if the readme content should be expanded a bit to explain difference?

spiiroin (approved):
Techincally LGTM.

Whether that readme update is needed is a matter of opinion.

okodron (approved):

permissions/Audio.permission
LaakkonenJussi: That was my original point and I did erase the README.md part. Well, I'll add it back then.

Tomin1: I guess it's okay as long as we clearly say that "ok, this gives you access to both (Audio = playback + recording) but you should really include also the other one (Microphone) if you need that functionality (recording)". IMO we should leave the technical bits of that out of the table because there just isn't a whole lot of space there.

LaakkonenJussi: @spiiroin Added more descriptions to `README.md` with eae11d6 - what do you think, too much or just enough?

README.md
```diff
@@ -111,7 +111,7 @@ Permissions that applications may use (names are subject to change):
 | Location | Use GPS and positioning. |
 | MediaIndexing | Access to Tracker to list files on device. If you have access to a data directory, you may want to use also this. |
 | Messages | Access to message data and to send SMS messages. |
-| Microphone | Record audio with microphone. Use Audio permission for playback of the recorded audio. |
+| Microphone | Record audio with microphone. Use Audio permission for playback of the recorded audio (but since Pulseaudio streams cannot be separated this enables also audio playback, which is subject to change ). |
```
okodron, line 114:
```suggestion
| Microphone | Record audio with microphone. Use Audio permission for playback of the recorded audio (but since Pulseaudio streams cannot be separated this enables also audio playback, which is subject to change). |
```

LaakkonenJussi: Removed the extra space with c1cc3d824a657a2ad0ed329bd3cb0d987cd5c1e0 (small change so force pushed).

permissions/Audio.permission
spiiroin: I think it is ok now.

PR #76, opened 2021-05-20
Abstract Telepathy access for messages.

LaakkonenJussi:
Abstract the Telepathy access for messages by using organization and
application names in the D-Bus name ownership. These are supported via
--template in sailjail that will get the org and app names passed and
processed by firejail.

### DIFF
--- a/permissions/jolla-messages.profile
+++ b/permissions/jolla-messages.profile
@@ -11,7 +11,5 @@
 ### APPLICATION
 # FIXME: This is a legacy D-Bus service name, we need to drop it at some point
 dbus-user.own org.nemomobile.qmlmessages
-# FIXME: another systematic dbus-name to deal with?
-dbus-user.own org.freedesktop.Telepathy.Client.qmlmessages
 
 dbus-user.call com.jolla.settings=com.jolla.settings.ui.showPage@/com/jolla/settings/ui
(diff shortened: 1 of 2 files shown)


LaakkonenJussi:
@okodron also, review is appreciated.

okodron (approved):
Ok in general. Needs https://github.com/sailfishos/firejail/pull/12 and https://github.com/sailfishos/sailjail/pull/42

permissions/jolla-messages.profile
```diff
@@ -11,7 +11,6 @@
 ### APPLICATION
 # FIXME: This is a legacy D-Bus service name, we need to drop it at some point
 dbus-user.own org.nemomobile.qmlmessages
-# FIXME: another systematic dbus-name to deal with?
-dbus-user.own org.freedesktop.Telepathy.Client.qmlmessages
+dbus-user.own org.freedesktop.Telepathy.Client.${OrganizationName}.${ApplicationName}
```
Tomin1, line 14: This is the rule we want but I'm not sure this is the place for it. Should we put this to Messages.permission?

LaakkonenJussi: Well, should we? Maybe we should as this is for all messaging apps to use? I'll make the change.

LaakkonenJussi: Moved with c1ca493

Tomin1 (approved):

dseight (approved):

PR #79, opened 2021-05-28
Add booster profile files

spiiroin:
Boosting sandboxed applications works by launching booster in sandbox.
Such application boosters use permissions for the the application they
are handling. Any additional sandboxing tweaks that are needed are
handled by use of booster type specific profile files.

### DIFF
--- a/permissions/booster.profile
+++ b/permissions/booster.profile
@@ -1,0 +1,25 @@
+# -*- mode: sh -*-
+
+# Single-instance booster feature needs to be
+# able to execute a helper binary.
+private-bin single-instance
+
+# Provide read-access to legacy privileges config
+whitelist /usr/share/mapplauncherd/privileges
+whitelist /usr/share/mapplauncherd/privileges.d
+
+# Some boosters are executed as systemd units
+# and need to have access to systemd socket for
+# sending startup notifications.
+noblacklist ${RUNUSER}/systemd
+read-only   ${RUNUSER}/systemd
+
+# Boosters needs to be able to create sockets.
+# Override read-only from Base.permission.
+ignore read-only ${RUNUSER}/mapplauncherd
+
+# Boosters need to be able to prompt user for launch
+# permission and query application information.
+# Allow ipc with sailjail daemon.
+dbus-system.talk org.sailfishos.sailjaild1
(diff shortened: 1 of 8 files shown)


Tomin1:
Just a thought, but should we have these profile files for applications separate from profile files for boosters? I application binaries and booster binaries are stored separately (_/usr/bin_ vs. _/usr/libexec/mapplaunched_). Then again it might not be a real issue as apps should not be prefixed with "booster-".

I assume these booster permissions don't get dropped when the app is executed, so does this open some holes like launching apps from a sandbox? That would basically mean that until that's fixed one should be careful which sandboxed apps get boosted.

Tomin1 (approved):
This looks fine as long as we have a plan to take care of any possible holes this might introduce.

okodron (approved):

spiiroin:
> This looks fine as long as we have a plan to take care of any possible holes this might introduce.

As it is not possible to modify sandbox once it is running, applications started via boosted sandbox will always have a bit wider view to system than the same app without boosting. That just needs to be taken in account when deciding whether boosting is done or not along with other details such as increased memory use etc. That said:

- Booster related data / helper binaries under /usr remain readable, but that should not be an issue
- We could perhaps already now limit permission daemon ipc to promp + get info
- Once we have those macros available it should be possible / relatively easy to limit booster socket directory access so that boosted application has write permissions only to the socket that affects the boosted application itself

permissions/booster.profile
```diff
@@ -0,0 +1,25 @@
...
+# sending startup notifications.
+noblacklist ${RUNUSER}/systemd
+read-only   ${RUNUSER}/systemd
+
+# Boosters needs to be able to create sockets.
+# Override read-only from Base.permission.
+ignore read-only ${RUNUSER}/mapplauncherd
+
+# Boosters need to be able to prompt user for launch
+# permission  and query applicaiton information.
```
Tomin1, line 22: Btw, some typos here.

spiiroin:
Rebased

permissions/booster.profile
spiiroin: Should be fixed

Tomin1 (approved):

spiiroin:
Review commits squashed

PR #80, opened 2021-05-31
Remove user dir includes from Email perms.

LaakkonenJussi:
The Email application should have the necessary permissions listed that
are then visible for the user. The includes used here wouldn't be shown.


LaakkonenJussi:
@okodron review is appreciated.

okodron (approved):

Tomin1 (approved):

PR #81, opened 2021-05-31
[permissions] Whitelist user's dconf. JB#54408 OMP#JOLLA-162

rainemak:


rainemak:
@okodron review appreciated.

okodron (approved):
OMP#JOLLA-162 as well?

Tomin1:
I wonder if whitelisting a file can cause bugs with older devices as they can't delete or rename a file that's mounted to another namespace (bug in older kernels). Could we whitelist the directory instead?

rainemak:
> I wonder if whitelisting the files can cause bugs with older devices as they can't delete or rename a file that's mounted to another namespace (bug in older kernels). Could we whitelist the directory instead?

Good idea. Adding whitelisting for the dirs + added mkdir just in case.

Tomin1 (approved):

PR #83, opened 2021-06-02
Add default profile configuration

Tomin1:
Default profile defines the set of permissions that is applied when
application doesn't define any application profile for sandboxing. It
allows sandboxing of applications antedating sandboxing.

See also: https://github.com/sailfishos/sailjail/pull/50


okodron (approved):

rainemak (approved):
Approving

dseight (approved):

PR #86, opened 2021-06-11
[permissions] Fix applications not receiving dconf notifications. Fixes JB#54428

adenexter:
DConf notifications are delivered on the ca.desrt.dconf.Writer interface
rather than simply ca.desrt.dconf.

### DIFF
--- a/permissions/Base.permission
+++ b/permissions/Base.permission
@@ -112,8 +112,8 @@ dbus-user.broadcast *=org.freedesktop.DBus.ObjectManager@/*
 # END sessionbus-filter.resource
 # BEG sessionbus-ca.desrt.dconf.Writer.resource
 dbus-user.talk ca.desrt.dconf
-dbus-user.call ca.desrt.dconf=ca.desrt.dconf.*@/*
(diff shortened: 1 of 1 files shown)


adenexter:
I suspect I'm missing something because the framing comment uses the name ca.desrt.dconf.Writer but the rules don't. Still I'm seeing value changes come through to applications where they weren't before.

Tomin1:
> I suspect I'm missing something because the framing comment uses the name ca.desrt.dconf.Writer but the rules don't. Still I'm seeing value changes come through to applications where they weren't before.

Looks like `xdg-dbus-proxy` is a little deceptive when it comes to interface rules: It seems that while `.*` can match dots in bus names, it can't match other parts of interface name than the method (or signal) at the end. This only talks about methods but I think it applies to signals as well:
https://github.com/flatpak/xdg-dbus-proxy/blob/15d2803039b1964dc95112d6d1624a291d338c4e/flatpak-proxy.c#L517

This difference is not explained on the man page.

Are there other interfaces in `ca.desrt.dconf` that would be interesting here or just that `ca.desrt.dconf.Writer`?

Tomin1 (approved):

PR #90, opened 2021-06-18
[permissions] Allow browser to open settings. Fixes JB#54522 OMP#JOLLA-187

mkenttala:


okodron (approved):

dseight (approved):

permissions/sailfish-browser.profile
```diff
@@ -11,6 +11,7 @@
 ### APPLICATION
 dbus-user.own org.sailfishos.browser
 dbus-user.own org.sailfishos.browser.ui
+dbus-user.talk com.jolla.settings
```
Tomin1, line 14: If there is some specific interface, path or both that browser calls, I would specify also that here with a dbus-user.call rule. Since it obviously wasn't using com.jolla.settings for anything else, it should be safe to add one.

okodron (approved):
Even better now.

Tomin1 (approved):

PR #95, opened 2021-07-02
Readme improvements

Tomin1:
Escape < and >. Otherwise Github interprets these as invalid HTML tags.

Update Exec line documentation. It's no longer necessary to define sailjail command on Exec line.

Use capitalized name for ApplicationName example. This is the convention so better to use it in examples too.


okodron (approved):

dseight (approved):

rainemak (approved):
LGTM

PR #97, opened 2021-07-12
Add profile for voicecall-ui-prestart.

Tomin1:
Add profile for voicecall-ui-prestart because it has different name from
the binary name. As it is the same application as voicecall-ui, use a
symbolic link.

Also remove extra stuff from voicecall-ui. Now that data directory is
set properly, we can also remove that FIXME and the whitelist line.


okodron (approved):

dseight (approved):

llewelld (approved):
Tested and it fixes the D-Bus issues experienced without the patch. LGTM

PR #98, opened 2021-08-06
[permissions] Add os-release symlink targets. JB#55070 OMP#OS-8462

dseight:


spiiroin (approved):
As it seems that we might get fix for underlying problem from firejail upstream, I think this is ok as workaround while waiting for that.

okodron (approved):

PR #99, opened 2021-08-09
[permission] Add Microphone permission to the Camera. Fixes JB#55078

rainemak:
Audio permission is needed in order to get microphone access.


rainemak:
This should be actually done in the app side.

PR #101, opened 2021-08-19
[permissions] Give sailfish-browser access to the device lock Authenticator interface. Contributes to JB#55160

adenexter:


permissions/sailfish-browser.profile
```diff
@@ -12,6 +12,9 @@
 dbus-user.own org.sailfishos.browser.ui
 dbus-user.call com.jolla.settings=com.jolla.settings.ui.showTransfers@/com/jolla/settings/ui
 
+dbus-system.talk org.nemomobile.devicelock
+dbus-system.call org.nemomobile.devicelock=org.nemomobile.devicelock.Authenticator.*@/*
```
adenexter, line 16: Technically it only needs access to the fixed /authenticator path, but that resulted in an invalid rule.

spiiroin: Interesting. The rule looks perfectly fine to me. And proxy would accept it, no complaints from:

    xdg-dbus-proxy 3>/dev/null --fd=3 fooname barsocket \
      --call='org.nemomobile.devicelock=org.nemomobile.devicelock.Authenticator.*@/authenticator'

But firejail does not like it (and thus it does not matter whether proxy would grok it or not):

    firejail --dbus-system.call='org.nemomobile.devicelock=org.nemomobile.devicelock.Authenticator.*@/authenticator'
    Error: Invalid dbus-system.call rule: org.nemomobile.devicelock=org.nemomobile.devicelock.Authenticator.*@/authenticator

After taking a peek at firejail check_object_path(), it seems that it does not allow object paths like "/foo", it needs to be either "/" or "/foo/bar[/baf]..." -> to me this looks like a bug in our current firejail version. @Tomin1 Any chance that it would have been fixed in upstream?

https://github.com/sailfishos-mirror/firejail/blob/b818fc40cef9865b3202b7de227348dd26acf9fd/src/firejail/dbus.c#L114

Tomin1: Yes, they have fixed it: https://github.com/sailfishos-mirror/firejail/commit/437be33f400e26858f398e021f1c896f132897eb

And it will be in 0.9.66. :smiley:

Tomin1: It should work now. That update was merged yesterday.

Tomin1 (approved):

PR #104, opened 2021-09-01
[permissions] Fix libsignond bus permission. JB#55409

HandsomeJaack:
com.google.code.AccountsSSO.SingleSignOn is located on system bus.
Fixes error: `SignOn::ConnectionManager::init:130 - Unable to activate p2p signond service!` when attempting to create email-account.

### DIFF
--- a/permissions/Accounts.permission
+++ b/permissions/Accounts.permission
@@ -12,9 +12,11 @@ whitelist /usr/share/accounts
 whitelist /usr/share/libsailfishkeyprovider/storedkeys.ini
 
 # To create accounts using libsignon
+mkdir ${RUNUSER}/signond
 whitelist ${RUNUSER}/signond
 read-only ${RUNUSER}/signond
 # Application must be able to create a socket here
+mkdir ${RUNUSER}/signonui
 whitelist ${RUNUSER}/signonui
 
 mkdir     ${PRIVILEGED}/Accounts


okodron (approved):
Works now.

pvuorela:
What i gather the service should be on session bus?

HandsomeJaack:
@pvuorela actually it's is, but there is [patch](https://github.com/sailfishos/libsignon/blob/master/rpm/0006-Always-use-P2P-DBus-if-enabled.-Contributes-to-JB-42.patch#L85) in libsignond which enables p2p by default, seems that firejail treats with p2p like system permission (but I haven't found clear evidence of this).

pvuorela:
If the system here somehow makes the /run/... socket readable it sounds more like accident than a real fix. Also the file here includes already 
```
# Application must be able to create a socket here
whitelist ${RUNUSER}/signonui
```
which I would expect to take care of the p2p socket. 

cc @Tomin1

Tomin1:
Yes, indeed. This change should not have any effect on those peer-to-peer sockets since dbus-system and dbus-user rules affect only the proxies on these two well-known buses. There is no proxy on peer-to-peer sockets. Access to those is controlled by file system rules like whitelist and read-only.

spiiroin:
> # Application must be able to create a socket here
> whitelist ${RUNUSER}/signonui

What creates that directory / sockets within it, and when?

IIRC for whitelisting to have any effect: the directory needs to already exist before sandbox is set up, or it needs to be created during sandbox setup.

Basically: should there be "mkdir ${RUNUSER}/signonui" rule before that whitelist rule?

HandsomeJaack:
Ok, I put together a picture, thank you all
@spiiroin I've check out your suggestion and it works, thanks

okodron (approved):

Tomin1 (approved):

PR #105, opened 2021-09-02
media: Give access to /etc/gst-droid to media and camera applications

abranson:
[media] Give access to /etc/gst-droid to media and camera applications. Fixes JB#55372

Config there enables optional codecs such as h265, and specifies camera quirks on some devices.

### DIFF
--- a/permissions/Base.permission
+++ b/permissions/Base.permission
@@ -72,6 +72,9 @@ private-etc localtime
 private-etc mtab
 private-etc mime.types
 private-etc kcapi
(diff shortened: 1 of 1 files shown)


Tomin1:
I would say that this rule could be in Base.permission even. Right now it looks like it's just information for libraries about video codes which you might need even without access to camera hardware or videos directory themselves. The library is anyway usable without the specific permissions but it's just that it won't work without this.

mlehtima (approved):

PR #106, opened 2021-09-07
[permissions] Allow Sharing by default. Fixes JB#55380

spiiroin:
Include Sharing.permissions from Base.permissions.

Update documentation so that "Sharing" is listed under permissions
that applications should not use directly.


pvuorela (approved):

Tomin1 (approved):

spiiroin:
Rebased

PR #108, opened 2021-10-15
[sailjail-permissions] Add a permission to talk over secrets daemon DBus. Contributes to JB#56300

dcaliste:
…Bus.

Allow GnuPG pinentry to talk to the secrets daemon also.

This requires sailfishos/sailfish-secrets#178 to be merged.
Without these two PRs, the email signature in email application is failing.

@chriadam and @pvuorela, I've no real idea if the dbus-user entries I've put here make sense or not. If you may review or ask anyone else to review, that would be great. Thank you.

### DIFF
--- a/permissions/Secrets.permission
+++ b/permissions/Secrets.permission
@@ -1,0 +1,16 @@
+# -*- mode: sh -*-
+
+# x-sailjail-translation-catalog =
+# x-sailjail-translation-key-description =
+# x-sailjail-description = Secrets
+# x-sailjail-translation-key-long-description =
+# x-sailjail-long-description = Use DBus API to access the Sailfish Secrets daemon
+
+mkdir     ${RUNUSER}/sailfishsecretsd
+whitelist ${RUNUSER}/sailfishsecretsd
+read-only ${RUNUSER}/sailfishsecretsd
+
+# Allow discovery of the p2p address used by the daemon.
+dbus-user.talk      org.sailfishos.secrets.daemon.discovery
+dbus-user.call      org.sailfishos.secrets.daemon.discovery=org.sailfishos.secrets.daemon.discovery.peerToPeerAddress@/Sailfish/Secrets/Discovery
+
--- a/README.md
+++ b/README.md
@@ -156,6 +156,7 @@ Internal permissions that applications generally should not use directly:
 | FingerprintSensor |
 | Notifications |
 | PinQuery |
+| Secrets |
 | Sensors |
 | Sharing |
 | Thumbnails |
--- a/permissions/GnuPG.permission
+++ b/permissions/GnuPG.permission
@@ -13,3 +13,8 @@ noblacklist ${HOME}/.gnupg
 private-bin gpg2
 private-bin gpg-agent
 private-bin pinentry
+
+# pinentry is talking to the Secrets daemon in pipe mode
+# this include can be safely removed when the gpg-agent will
+# run in daemon mode.
+include /etc/sailjail/permissions/Secrets.permission


permissions/Secrets.permission
```diff
@@ -0,0 +1,17 @@
...
+# x-sailjail-translation-key-long-description =
+# x-sailjail-long-description = Use DBus API to access the Sailfish Secrets daemon
+
+mkdir     ${RUNUSER}/sailfishsecretsd
+whitelist ${RUNUSER}/sailfishsecretsd
+read-only ${RUNUSER}/sailfishsecretsd
+
+# Allow discovery of the p2p address used by the daemon.
+dbus-user.talk      org.sailfishos.secrets.daemon.discovery
+dbus-user.call      org.sailfishos.secrets.daemon.discovery=org.sailfishos.secrets.daemon.*@/*
```
Tomin1, line 15: Could we limit the methods or paths that are available or is this such interface that we can allow everything?

I know there are other places that we use full wildcards but in general we should pay more attention to these.

Tomin1 (commented):
Look mostly good to me. I recommed checking the exposed methods in that D-Bus interface and decide how much of them we want to expose. It might be that it's fine as is too.

I would also add this to README to the internal permissions list. It's correct to have an empty translation catalog for a permission that doesn't need translations as it's internal (e.g. only used as part of some other permission).

permissions/Secrets.permission
dcaliste: The only method that is required is defined like that in the code:
```
    QDBusInterface iface("org.sailfishos.secrets.daemon.discovery",
                         "/Sailfish/Secrets/Discovery",
                         "org.sailfishos.secrets.daemon.discovery",
                         QDBusConnection::sessionBus());
    if (iface.isValid()) {
        QDBusReply<QString> reply = iface.call("peerToPeerAddress");
```
I'm not an expert in DBus archanes and not at all in firejail rules about DBus but after reading a bit the manual, I guess the two rules are fine:
```
dbus-user.talk      org.sailfishos.secrets.daemon.discovery
dbus-user.call      org.sailfishos.secrets.daemon.discovery=org.sailfishos.secrets.daemon.discovery@/Sailfish/Secrets/Discovery
```

I've updated the PR accordingly.

dcaliste:
I've added `Secrets` to the list in the README of permissions not to be directly used.

I've also commented in the GnuPG permission that the secret permission inclusion there can be safely removed once the gpg-agent will run in daemon mode.

permissions/Secrets.permission
Tomin1: [Firejail documentation](https://firejail.wordpress.com/features-3/man-firejail-profile/) is a bit terse on this but [xdg-dbus-proxy](https://manpages.debian.org/stretch-backports/xdg-dbus-proxy/xdg-dbus-proxy.1.en.html) explains these a bit better. Xdg-dbus-proxy is what firejail will ultimately use and although firejail also parses those rules they are passed to xdg-dbus-proxy in the same format. So for `talk` you need to use the service's bus name which is correct here, and then for `call` you need to have \<bus name\>=\<method name\>/\<path of the object\> that needs a little bit adjusting now. In this case `call` rule should be like this:
```
dbus-user.call      org.sailfishos.secrets.daemon.discovery=org.sailfishos.secrets.daemon.discovery.peerToPeerAddress@/Sailfish/Secrets/Discovery
```
Similarly for `broadcast` you can limit it to certain signals by specifying the signal names (using multiple rules if needed).

dcaliste: Ah, thanks for the explanation. It makes a bit more sense for me now.

I'm still surprised that my previous rule, without the proper method name, was actually working. In fact, even removing the call rule completely still works…

Tomin1: Well, `talk` rule without `call` rule would work because then you don't limit the methods you can make a call to.

I'm not sure how the version you have in the comment above could work. Are you sure it was sandboxed? I didn't try it myself though.

dcaliste: Yes, here is my test case (slightly unrealistic, but easy to use) :
```
QT_LOGGING_RULES="org.sailfishos.*=true" /usr/bin/sailjail /usr/bin/gpg2 -s ~/Documents/toto
```
with:
```
[nemo@Sailfish devel]$ cat /usr/share/applications/gpg2.desktop 
[Desktop Entry]
Type=Application
Name=Mail
Icon=icon-launcher-email
Exec=/usr/bin/gpg2 -s /home/nemo/Documents/toto
X-MeeGo-Logical-Id=mail-ap-name
X-MeeGo-Translation-Catalog=jolla-email
X-Maemo-Service=com.jolla.email.ui
X-Maemo-Method=com.jolla.email.ui.activateWindow
X-Maemo-Object-Path=/com/jolla/email/ui

X-Desktop-File-Install-Version=0.26

[X-Sailjail]
Permissions=Accounts;Contacts;Phone;Email;WebView;Privileged;Internet;AppLaunch;Calendar;UserDirs;RemovableMedia;MediaIndexing;Sharing
OrganizationName=com.jolla
ApplicationName=email
```
And the PR moving the sailfish secret socket applied.

When working, the output starts with:
```
[D] unknown:0 - Connecting to secrets daemon via p2p address: "unix:path=/run/user/100000/sailfishsecretsd/p2pSocket" with connection name: "sailfishsecretsd-connection-0"
[D] unknown:0 - Connected to secrets daemon via connection: "sailfishsecretsd-connection-0"
```
When not working, it starts with :
```
[D] unknown:0 - Unable to query the peer to peer socket address from the secrets daemon!  Using fallback address.
```

chriadam:
@Tomin1 what are your thoughts on this one, can we get it in (along with https://github.com/sailfishos/sailfish-secrets/pull/178/files)?

Tomin1 (approved):
> @Tomin1 what are your thoughts on this one, can we get it in (along with https://github.com/sailfishos/sailfish-secrets/pull/178/files)?

Yes, both look good to me.

(Last time I was reviewing this, I got lost in trying to replicate the results dcaliste had in the discussion but I think I didn't get far enough to generate a key for myself to try it. It still seems strange to me that the version of the rules which doesn't specify method (or wildcard) would work somehow. The rules, however, look correct now.)

PR #111, opened 2021-11-04
Add access to policies

Tomin1:
Add access to MDM policies for those apps that need it.

Drop metadata lines from profile files. These were never used. It makes sense to clean them up.

Rename booster.profile to booster.inc. This makes more sense and avoids confusion. This .inc extension is already used by firejail for this very same purpose.


pvuorela:
Should we have some common file included in these places to avoid duplicating the rules?

Tomin1:
> Should we have some common file included in these places to avoid duplicating the rules?

I had the same thought but thought that we didn't use includes in .profile files. I was actually wrong about that as we have fairly recently added booster.profile which is included by all booster-*.profile files. However that profile ending is a bit problematic, I think I'll use .inc which is already used by firejail.

Tomin1:
> Should we have some common file included in these places to avoid duplicating the rules?

Fixed. Also changed that booster.profile to booster.inc.

pvuorela (approved):

spiiroin:
> However that profile ending is a bit problematic, I think I'll use .inc which is already used by firejail.

Now, if we would just need rename those "these are not really permission files" to inc too, we'd be close to how things were initially set up ;-)

PR #113, opened 2021-11-17
Add Compatibility permission

Tomin1:
Add Compatibility permission for default profile. This allows apps to
access selected binaries from /usr/bin. It has been used as a way around
certain library restrictions.

Please check if the list of binaries is sane and if it should have something
more or less. Also please check the descriptions.

See also https://github.com/sailfishos/sailjail/pull/69

### DIFF
--- a/permissions/Compatibility.permission
+++ b/permissions/Compatibility.permission
@@ -1,0 +1,32 @@
+# -*- mode: sh -*-
+# This provides compatibility for older apps
+# The permission is not shown to user,
+# thus it ***must not*** give access to additional data or hardware resources
+
+# x-sailjail-translation-catalog =
+# x-sailjail-translation-key-description = permission-la-compatibility
+# x-sailjail-description = Compatibility
+# x-sailjail-translation-key-long-description = permission-la-compatibility_description
+# x-sailjail-long-description =
+
+# PDF tools
+private-bin pdfinfo
+private-bin pdftocairo
+private-bin pdftoppm
+private-bin pdftops
+
+# Calligra for office file conversion
+private-bin calligraconverter
+
+# ffmpeg
+private-bin ffmpeg
+private-bin ffprobe
+
+# Package manager
+# Note that this does not allow install or removal
+private-bin rpm
+
+# Basic tools
+private-bin busybox
+private-bin tar
+private-bin which
--- a/README.md
+++ b/README.md
@@ -151,6 +151,7 @@ Internal permissions that applications generally should not use directly:
 | :---       | :---        |
 | Base       | Base set of permissions that every application is granted implicitly. |
 | CaptivePortal |
+| Compatibility | Access to binaries for compatibility. Direct use of binaries should be avoided. |
 | Connman |
 | GnuPG |
 | FingerprintSensor |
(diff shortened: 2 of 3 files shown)


permissions/Compatibility.permission
```diff
@@ -0,0 +1,41 @@
...
+# The permission is not shown to user,
+# thus it ***must not*** give access to additional data or hardware resources
+
+# x-sailjail-translation-catalog =
+# x-sailjail-translation-key-description = permission-la-compatibility
+# x-sailjail-description = Compatibility
+# x-sailjail-translation-key-long-description = permission-la-compatibility_description
+# x-sailjail-long-description =
+
+# PDF tools
```
attah, line 12: Nice!
I don't want to be presumptuous... but i presume this is mostly for SeaPrint?
In that case i can tell you i only use pdftoppm, pdftops, pdftocairo, pdfinfo currently (SeaPrint 0.9).
For 1.0 i am dropping the use of all poppler-utils binaries, so i can't see that list growing at all.
But this is really a good respite from having to get that completely sorted with a deadline approaching.

permissions/Compatibility.permission
```diff
@@ -0,0 +1,41 @@
...
+private-bin pdfsig
+private-bin pdftocairo
+private-bin pdftohtml
+private-bin pdftoppm
+private-bin pdftops
+private-bin pdftotext
+private-bin pdfunite
+
+# Calligra for office file conversion
+private-bin calligraconverter
```
attah, line 28: For calligraconverter however, i don't see any alternatives in the short term, or that i could do independently.
In the longer term, perhaps sailfish-office can grow a share-as-pdf functionality, or even more formal print functionality...

permissions/Compatibility.permission
```diff
@@ -0,0 +1,41 @@
...
+# thus it ***must not*** give access to additional data or hardware resources
+
+# x-sailjail-translation-catalog =
+# x-sailjail-translation-key-description = permission-la-compatibility
+# x-sailjail-description = Compatibility
+# x-sailjail-translation-key-long-description = permission-la-compatibility_description
+# x-sailjail-long-description =
+
+# PDF tools
+private-bin pdfattach
```
attah, line 13: pdfattach is perhaps the one that could have the most nefarious use cases, but nothing that couldn't be replicated.

attah (commented):
This is great news :)

poetaster (commented):
ffmpeg and ffprobe are great news for me. Both Videoworks and the Audiocut app depend on them and have used custom compiled versions to date. In testing, installing ffmpeg-tools sufficed so it could make the pkgs a lot smaller and depend on SFOS audited versions.

permissions/Compatibility.permission
Tomin1: Yes. I don't know anything else that would be using these tools. I included everything initially but the idea was to reduce them before merging depending on what's really needed. Thanks for telling me know which tools you actually use, I appreciate that!

vigejolla (approved):
LGTM

rainemak (approved):

dseight (approved):

okodron (approved):

PR #114, opened 2021-11-30
[permissions] Add /etc/media_codec.xml to base permissions. JB#54037

mlehtima:
Some device load codec configuration from /etc.


spiiroin (approved):
Might cause cosmetic issues (warnings in device that do not have this file), but atm there is little that can be done about it -> LGTM

PR #115, opened 2021-12-02
[permissions] Be more verbose about avoiding Compatibility permission

vigejolla:


Tomin1:
Commit message seems to have split to multiple lines. JB#... needs to be on the same line.

README.md
```diff
@@ -151,7 +151,7 @@ Internal permissions that applications generally should not use directly:
 | :---       | :---        |
 | Base       | Base set of permissions that every application is granted implicitly. |
 | CaptivePortal |
-| Compatibility | Access to binaries for compatibility. Direct use of binaries should be avoided. |
+| Compatibility | Access to binaries for compatibility. Direct use of binaries should be avoided. Please note that this permission is provided in order to support old applications, new applications should try to use other means. The binaries may disappear in any release, especially if we provide APIs with similar functionality. |
```
Tomin1, line 154: If wonder if this should also (explicitly) say "for any reason".

```suggestion
| Compatibility | Access to binaries for compatibility. Direct use of binaries should be avoided. Please note that this permission is provided in order to support old applications, new applications should try to use other means. The binaries may disappear in any release, for any reason, but especially if we provide APIs with similar functionality. |
```

Tomin1 (approved):
One comment you might want to agree or disagree with but I think this LGTM either way.

PR #118, opened 2022-01-28
[permissions] Supplement broadcast permissions for UDisksListen. OMP#OS-5687

DieTime:
For the correct synchronization of the partition model, it is necessary to enable the UDisks2 service to listen to the signals of  `org.freedesktop.DBus.ObjectManager.*` and `org.freedesktop.DBus.Properties.*` interfaces.

Without them, the status of the partition is synchronized incorrectly (mounted or unmounted, etc.).

### DIFF
--- a/permissions/UDisksListen.permission
+++ b/permissions/UDisksListen.permission
@@ -7,6 +7,8 @@
 # x-sailjail-long-description = Permissions to access properties and to listen to property changes on storage devices
 
 # BEG systembus-org.freedesktop.UDisks2.resource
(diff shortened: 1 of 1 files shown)


aroditelev (approved):

spiiroin (changes requested):
Rationale and changes make sense to me.

However, the explanation is only in PR discussion part - which is not readily available when browsing git history later on -> I'd suggest adding it also to commit message.

DieTime:
Added a description to the commit message

dseight (approved):

spiiroin (approved):

rainemak (approved):

PR #120, opened 2022-02-16
[permissions] Allow browser to receive dsme dbus signals

mkenttala:
Sailfish browser needs to get dsme reboot and shutdown signals.


permissions/sailfish-browser.profile
```diff
@@ -5,6 +5,8 @@ dbus-user.call com.jolla.settings=com.jolla.settings.ui.showTransfers@/com/jolla
 dbus-system.talk org.nemomobile.devicelock
 dbus-system.call org.nemomobile.devicelock=org.nemomobile.devicelock.Authenticator.*@/authenticator
 
+dbus-system.talk com.nokia.dsme
```
Tomin1, line 8: Could we have a comment here why this is needed? Also I would like to see this limited to those signals that are actually needed by adding some broadcast rules instead of allowing calls to everything. So something to this direction (I didn't check if these are the correct signals in this instance, it might be that `com.nokia.dsme.signal.*` is the right choice, but not necessarily):
```
dbus-system.broadcast com.nokia.dsme=/com/nokia/dsme/signal@com.nokia.dsme.signal.shutdown_ind
dbus-system.broadcast com.nokia.dsme=/com/nokia/dsme/signal@com.nokia.dsme.signal.state_change_ind
```

Tomin1 (commented):
Also commit message is missing a bug reference.

permissions/sailfish-browser.profile
spiiroin: Yes, it might be informative to know what signals browser needs. But...

... in all likelihood access to dsme could/should be given in *Base*. Something like all of signals some and at least getters from methods. 

Some/most/all of the signals are meant for application consumption e.g. shutdown / save_unsaved_data signals ref https://wiki.maemo.org/Documentation/Maemo_5_Developer_Guide/Application_Development/LibOSSO_library

Also dsme already ignores e.g. reboot requests from unprivileged clients. If we want to block also attempts to make such requests, applications probably should be denied access also to dsme specific unix socket ipc too (via libdsme) - which would deny access also to process watchdog feature etc.

spiiroin (changes requested):
All applications should be able to receive notifications related to device state changes -> relaxing rules should happen in Base.permission if possible.

permissions/sailfish-browser.profile
mkenttala: Moved signals rule to Base.permission.

spiiroin (approved):

Tomin1 (approved):

rainemak (approved):
Please rebase upgrade-4.4.0 on top of master as well. Trigger build for both.

PR #121, opened 2022-04-03
[permissions] Add /var/run to base permissions.

daniellinjama:
Connman creates /etc/resolv.conf as a symbolic link to /var/run/connman/resolv.conf.

### DIFF
--- a/permissions/Base.permission
+++ b/permissions/Base.permission
@@ -58,6 +58,7 @@ include whitelist-usr-share-common.inc
 whitelist /var/lib/timed
 blacklist /var/lib/timed/shared_events
 blacklist /var/lib/timed/shared_settings
+whitelist /var/run
 # FIXME: /var handling needs more thought
 
 # /etc:


daniellinjama:
Issue reported on Sailfish OS Forum: https://forum.sailfishos.org/t/10829

rainemak:
Thank you @daniellinjama for your contribution. Will this allow you to run "dnsmasq"?

daniellinjama:
> Thank you @daniellinjama for your contribution. Will this allow you to run "dnsmasq"?

Yes, this allows running dnsmasq without breaking resolver configuration inside sailjail. I have tested this change on Xperia 10 Plus running Sailfish OS 4.4.0.58

Thaodan:
`/run`  should also be added or just alone that. `/var/run` is the legacy path.

spiiroin:
> `/run` should also be added or just alone that. `/var/run` is the legacy path.

Actually, /var/run should not be used at all - especially with anything related to sandboxing as it just causes issues. We've already patched number of services with this in mind and seems that this particular connman file slipped through.

Also, whitelisting too-close-to-root directories should be avoided. If truly needed, whitelisting relevant topic specific subdirectory might be an option.

But in this case it should not be needed as sandboxed applications already see /run/connman/resolv.conf and the problem is use of legacy path /var/run/connman/resolv.conf. As connman sets up that symlink -> IMO this is bug in connman, not something to fix in permissions. @LaakkonenJussi 

```
# firejail --ls=$(pidof sailfish-browser) /run/connman/resolv.conf
-rw-r--r-- root     root              70 resolv.conf
# firejail --ls=$(pidof sailfish-browser) /var/run/connman/resolv.conf
Error: cannot access /var/run/connman/resolv.conf
```

Thaodan:
> > `/run` should also be added or just alone that. `/var/run` is the legacy
> > path.
> Actually, /var/run should not be used at all - especially with anything
> related to sandboxing as it just causes issues. We've already patched
> number of services with this in mind and seems that this particular connman
> file slipped through.
Well that's an issue in firejail and the kernel. Patching every service to 
avoid /run or /var/run just backfires..

spiiroin:
> Patching every service to avoid /run or /var/run just backfires..

Using /run is ok. Using /var/run is not. And afaict only connman remains referring to /var/run.

spiiroin:
> And afaict only connman remains referring to /var/run.

Bah, naturally looking for /var/run/connman yielded hits only in connman ;-)

But in the context of sandboxing the thing that matters are symlinks that go through /var/run, and of those I found only /etc/resolv.conf -> /var/run/connman/resolv.conf

daniellinjama:
Created pull request to connman https://github.com/sailfishos/connman/pull/26

PR #124, opened 2022-05-10
[sailjail-permissions] Add a voicecall-ui-prestart profile.

dcaliste:
This allows to declare the prestart voicecall-ui
as owner of the com.jolla.voicecall.ui
and com.nokia.telephony.callhistory interfaces.

This is related to https://forum.sailfishos.org/t/4-4-0-58-tap-missing-call-doesnt-open-phone-app/10819. The problem is arising from the fact that within sailjail, the prestart voicecall-ui cannot register the `com.nokia.telephony.callhistory` service because it's not allowed to own it.

I'm not sure this is the right way to solve the problem though it is efficient ;)

### DIFF
--- a/permissions/voicecall-ui-prestart.profile
+++ b/permissions/voicecall-ui-prestart.profile
@@ -1,0 +1,3 @@
+# -*- mode: sh -*-
+dbus-user.own com.jolla.voicecall.ui
+dbus-user.own com.nokia.telephony.callhistory


llewelld:
I'm not able to reproduce the issue here, and I'm wondering if it might have already been fixed by the changes described in #117? 
Essentially, the voicecall-ui-prestart.desktop file was dropped, so the voicecall-ui.profile is now used in all cases.

llewelld:
Just to clarify, the fixes that actually address the issue aren't in this repo: an updated version of the voicecall-ui app is needed for them.

dcaliste:
Ok, if it's fixed in voicecall-ui, then it's fine to close this PR.

llewelld:
@Tomin1 would you mind taking a look and confirming whether or not this is fixed please?

Tomin1:
> @Tomin1 would you mind taking a look and confirming whether or not this is fixed please?

It's broken in 4.4.0 because commit 26efe54b1fdcc63fc6e707f084544f3a43970947 was included there but a change to voicecall-ui was not. This must have been a mistake but I'm not sure how it happened. We either need to revert that commit here or include the change to remove voicecall-ui-prestart.desktop in voicecall-ui. Either should be safe to do.

llewelld:
> We either need to revert that commit here or include the change to remove voicecall-ui-prestart.desktop in voicecall-ui.

If I understand right the changes to drop voicecall-ui-prestart.desktop were already merged in, so perhaps it would make most sense to cherry-pick those changes?

Tomin1:
> > We either need to revert that commit here or include the change to remove voicecall-ui-prestart.desktop in voicecall-ui.
> 
> If I understand right the changes to drop voicecall-ui-prestart.desktop were already merged in, so perhaps it would make most sense to cherry-pick those changes to voicecall-ui?

That makes sense to me.

llewelld:
@Tomin1 has cherry-picked the changes in voicecall-ui that will fix this in the next release, so I'll close this PR. Thanks for the contribution @dcaliste which made the issue very clear and saved us a lot of time.

Tomin1:
> One question maybe @Tomin1 if you have time, since voicecall-ui sources are not public, I cannot check : is the missing commit in voicecall-ui changing the DBus service name for callhistory and putting under com.jolla.voicecall ? Or is it a completely different approach ?

So the issue here was that we have had a separate sandboxing profile (a desktop file, named `voicecall-ui-prestart.desktop`) for `voicecall-ui-prestart.service` and that determines which profile file Sailjail uses and that corresponding file was missing because the symlink was removed. The change in voicecall-ui changes it to always use the main file `/usr/share/applications/voicecall-ui.desktop` so that Saijail uses `voicecall-ui.profile` that already exists in the repository. No D-Bus names were changed for that.

dcaliste:
Ok, thanks.

PR #125, opened 2022-07-14
[docs] Advice on data migration. JB#58384

martyone:


README.md
```diff
@@ -69,17 +69,18 @@ application restarts.
 
 ### Use correct application data directories
 
-Modify Qt application main function by setting the organization and application name
-to match the declaration added to the desktop file.
+Applications that are not based on the Sailfish App library
+([libsailfishapp](https://github.com/sailfishos/libsailfishapp))
```
Thaodan, line 73: Would it not make more sense to say apps that are not using the Sailfish App library instead of being based on it?

README.md
```diff
@@ -69,17 +69,64 @@ application restarts.
...
+
+    const QString settingsPath =
+        QStandardPaths::writableLocation(QStandardPaths::AppConfigLocation)
+        + "/" + QCoreApplication::applicationName() + ".conf";
+    QSettings settings(settingsPath, QSettings::NativeFormat);
+
+Unless you are creating a new application, it is also necessary to take care about migrating
+application data and configuration from the old location outside the sandbox.
+
+An example of how `QSettings`-based configuration may be migrated:
```
rainemak, line 97: Migration on the spec & post install script would be a good example as well.

Thaodan: Those work on system level and don't act as the user. It's easier to do it as the running app.

README.md
```diff
@@ -69,17 +69,64 @@ application restarts.
...
+
+    if (!settings.contains("migrated")) {
+        const QString oldSettingsPath =
+            QStandardPaths::writableLocation(QStandardPaths::ConfigLocation)
```
Thaodan, line 101: You can use  `QStandardPaths::AppConfigLocation` here  instead of QStandardPaths::ConfigLocation and skip the first:
`+ "/" + QCoreApplication::applicationName()`.

martyone: That's the old config location. You can't use AppConfigLocation here as it uses the now set organization name.

README.md
```diff
@@ -69,17 +69,64 @@ application restarts.
...
+            + "/" + QCoreApplication::applicationName()
+            + "/" + QCoreApplication::applicationName() + ".conf";
+        QSettings oldSettings(oldSettingsPath, QSettings::NativeFormat);
+
```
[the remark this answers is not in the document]
martyone: Yup, I noticed your suggestion on Forum and actually I initially I liked it more than Ville's approach, but after all chose to go with Ville's as it feels a bit more versatile.

README.md
martyone: Fixed

README.md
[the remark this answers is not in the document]
Thaodan: Versatile why hat?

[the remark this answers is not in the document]
martyone: You can update the values as you copy them (some may contain paths), you can deal with possible failure (who knows when OOM killer steps in etc.) because successful migration is clearly determined by storing some config value at the end (with your approach you would have to copy files, update config then remove original files or so), etc.

README.md
rainemak: Would work for each. We have done it similarly on spec for some Jolla apps.

rainemak (approved):

Issue #127, opened 2022-08-02
[Documentation] How should "internal" permissions be avoided?


Olf0:
The header of the second table in [the section "Permissions" of the README currently states](https://github.com/sailfishos/sailjail-permissions/tree/6a79d43b20b1fcd70d7c4d56bf63fde574c91af5#permissions): "Internal permissions that applications generally should not use directly:"

The "internal permissions" listed cover many functions which are commonly in use (e.g., Notifications or Sharing), plus some functions which Jolla strongly promoted (e.g., Secrets and the again overhauled sharing API).

Hence my question is: How does Jolla envision to avoid using these "internal" permissions for third party apps, while keeping the extant functionality of an app intact, including the use of Notifications, Jolla's recent Sharing API and the Jolla-Secrets API?

I would appreciate much, if the documentation is updated to clarify this and to provide some guidance what app developers should do (because that it its purpose, IMO).

Side note: Most of the entries in the second column of the "internal permissions" table in [the section "Permissions" of the README currently states](https://github.com/sailfishos/sailjail-permissions/tree/6a79d43b20b1fcd70d7c4d56bf63fde574c91af5#permissions) are empty.  It is impossible to derive the exact scope of a SailJail permission only by its name, hence filling these voids would be very helpful, because one may over-interpret the scope of a permission without a description.

attah:
I'll attempt an explanation:
Not all are internal as in _internal use only_, but _library_ permissions included by ones that you are allowed to use.
Specifically Notifications and Sharing are definitely of that kind.
I.e. the vision is to (only have to) use derived permissions.

Compatibility was added for me, Poetaster and some others that got creative and skirted harbour rules with calling binaries in QProcess.
It really doesn't do more than allow access to some select binaries: https://github.com/sailfishos/sailjail-permissions/blob/master/permissions/Compatibility.permission

Olf0:
> I'll attempt an explanation:

Thank you for your explanations.  Actually they depict to bigger gaps in the documentation than I originally assumed when I created this issue.

> Not all are internal as in _internal use only_, but _library_ permissions included by ones that you are allowed to use. Specifically Notifications and Sharing are definitely of that kind.

IMHO "library permission" is a new term, which is not used anywhere else yet.  I think it is misleading, as it is seems to be completely unrelated to software libraries (i.e., a set of callable functions), e.g., shared objects / DLLs.

[...]

PR #128, opened 2022-08-04
[permissions] Update README to clarify which apps are sandboxed. Fixes JB#58455

### LINKED ISSUE #126: [Documentation] Unclear scope of "all apps"
The fourth sentence of [the Sailjail README](https://github.com/sailfishos/sailjail-permissions/tree/6a79d43b20b1fcd70d7c4d56bf63fde574c91af5#readme) currently states: "Target is to have application sandboxing enforced to all applications, starting from Sailfish OS 4.4.0."

* I wonder what "all" is supposed to mean precisely:
  1. … enforced for all applications by Jolla, …
  2. … enforced for all applications available at the Jolla Store ("harbour"), …
  3. … enforced for every application installed, …
  
  Please clarify the wording in the README, e.g., with one of my suggestions.

* Jolla obviously missed the goal of reaching this target by Sailfish OS 4.4.0.  AFAIU the Jolla Store app is not sandboxed, yet.
   While missing due dates is quite common in the IT-business, you might want to update the Sailfish OS release version in this sentence with your current target release.

llewelld:
The README suggested there was a plan to sandbox all apps by the
Sailfish OS 4.4.0 release. In practice, this meant that all apps without
permissions specified would be given a default profile from 4.4.0
onwards.

Since this is now the case, this change updates the README to reflect
the current situation.

Fixes #126.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,8 +3,9 @@
 Sailfish OS application sandboxing and permissions system is build on top of
 Firejail. The Firejail documentation can be found from [here](https://firejail.wordpress.com/).
 
-The Sailfish OS application permissions are applicable only to sandboxed applications. Target is to
-have application sandboxing enforced to all applications, starting from Sailfish OS 4.4.0.
+The Sailfish OS application permissions are applicable only to sandboxed applications. Since
+Sailfish OS 4.4.0 all applications that do not specifically opt-out or define permissions are
+sandboxed automatically using a default profile.
 
 Application launch happens via standard desktop file that has been augmented
 so that it contains necessary metadata (such as a list of requested permissions)


abranson (approved):

rainemak (approved):
This is good like this. Worth creating a task regarding Sharing permission and adjust some parts of the text. Sharing permission itself is internal while at the same being part of the Base permission. This implies that each sandboxed application will automatically gain Sharing permission as each app has base permissions.

Olf0 (approved):
Thank you @llewelld, your rephrased version of this central statement is concise and unambiguous (at least for me).

Olf0:
> … Worth creating a task regarding Sharing permission and adjust some parts of the text. Sharing permission itself is internal while at the same being part of the Base permission. This implies that each sandboxed application will automatically gain Sharing permission as each app has base permissions.

May I suggest not to add additional text somewhere, but fill out the empty field for the Sharing permission in then table of "internal permissions", as suggested in the side note of the original report at issue #127.

rainemak:
> > … Worth creating a task regarding Sharing permission and adjust some parts of the text. Sharing permission itself is internal while at the same being part of the Base permission. This implies that each sandboxed application will automatically gain Sharing permission as each app has base permissions.
> 
> May I suggest not to add additional text somewhere, but fill out the empty field for the Sharing permission in then table of "internal permissions", as suggested in the side note of the original report at issue #127.

You're failing to understand different aspect of this. What is allowed in Harbour is not everything and all should documented. I do understand that everything is not perfectly documented but we're working it. Sharing is implicitly allowed. Thus, should be documented as well.

Olf0:
> > May I suggest not to add additional text somewhere, but fill out the empty field for the Sharing permission in then table of "internal permissions", as suggested in the side note of the original report at issue #127.
> 
> You're failing to understand different aspect of this.

Do I?  Which one(s) are you thinking if?

> What is allowed in Harbour is not everything and all should documented.

While this is certainly true, this sentence does not make much sense IMO.  Especially as this issue, this discussion and the whole Sailjail permission system is unrelated to the Jolla Store ("Harbour") AFAIU.  Or am I missing something?

> I do understand that everything is not perfectly documented but we're working it. Sharing is implicitly allowed. Thus, should be documented as well.

I fully concur and have not stated anything contradicting.  But as you are explicitly replying to my suggestion to use the empty description field of the `Sharing` permission in the table "internal permissions" for this (instead of text somewhere else) this sound as if you want to contradict, but I fail to see any contradicting argument.

llewelld:
Rebased.

llewelld:
Thanks for all of the input. I've created a separate task related to the Sharing permission text. I'll try to take what you've said on board, but once there are some PRs related to it I'll post the links here and then I suggest it might be easiest to continue the conversation there.

llewelld:
I'm not sure whether I covered everything, but I've proposed some text about the Sharing permission and would welcome your comments:
https://github.com/sailfishos/sailjail-permissions/pull/129

PR #129, opened 2022-08-09
[permissions] Add Sharing permission details to the README. Fixes JB#58483

llewelld:
Adds some info about the Sharing permission for accessing the Share API,
specifically to clarify that it is available to all applications by
default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,7 +207,7 @@ Internal permissions that applications generally should not use directly:
 | Notifications |
 | PinQuery |
 | Sensors |
-| Sharing |
+| Sharing | Allows an application to access the [Share API](https://sailfishos.org/develop/docs/declarative-transferengine/), for sending data to other applications. The Share API is fully mediated and under the user's explicit control, and is therefore considered low risk. Consequently this permission is contained as part of the Base permission and hence implicitly granted to all applications.
 | Thumbnails |
 | UDisks | Permissions to call UDisks functions, includes UDisksListen. |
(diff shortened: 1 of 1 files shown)


Thaodan:
Can you mention the version when to are available? Like for example: "Available since version x.x"?

llewelld:
> Can you mention the version when to are available? Like for example: "Available since version x.x"?

The Share API and Sharing permission were both released in 4.2.0, after the sanboxing was made available in 4.0.1.

Thaodan:
> > Can you mention the version when to are available? Like for example: "Available since version x.x"?
> 
> The Share API and Sharing permission were both released in 4.2.0 (after the sandboxing was made available in 4.0.1). Given this, I'm thinking this kind of info might be more appropriate in the Share API documentation itself rather than here. What do you think?

Hm it might make sense to have it in both since it is quite useful to have it in this list too, I was thinking of another column that says  so.

Olf0:
> Hm it might make sense to have it in both since it is quite useful to have it in this list too, I was thinking of another column that says so for each permission.

Two separate places to maintain the same information is usually not a good idea, because the have to be concurrently updated.  Usually links are a much better solution, because they alleviate the issue of "updated info here, but missed to update its copy somewhere else".

Olf0 (approved):
The created description text is fine IMO.

Tomin1 (approved):

rainemak:
@llewelld could you rebase and merge?

llewelld:
Rebased. Sorry for the delay in getting this merged; I lost track of og.

PR #131, opened 2022-08-09
[permissions] Link permission names in documentation 

Olf0:
… (i.e., tables in README.md) to their definition.  Also sorted them strictly alphabetical (GnuPG ↔ FingerprintSensor) and checked tables for completeness (which was already the case).

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,55 +162,55 @@ Some assumptions about the application are made:
 
 Permissions that applications may use (names are subject to change):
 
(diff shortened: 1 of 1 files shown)


Olf0:
@llewelld, if this is of any interest, I might rebase it (as time permits, because that has too be done manually due to conflicts), if there is some indication that this MR will likely not stay unread until it is stale and "unmergeable", again.

PR #137, opened 2023-02-07
[sailjail-permissions] Make translation order reproducible. Fixes JB#60107

mlehtima:


pvuorela (approved):

tools/generate_translation_strings.py
```diff
@@ -65,8 +65,8 @@ def __str__(self):
 
 def generate(translations):
     print("/* Generated dummy file, do not edit */")
-    for file in filter(lambda name: name.endswith(".permission") or name.endswith(".profile"),
-                       os.listdir(translations)):
+    for file in sorted(list(filter(lambda name: name.endswith(".permission") or name.endswith(".profile"),
```
Tomin1, line 68: I don't think `sorted()` requires list type. It takes iterable, so `list()` call is redundant.

Also update indentation on the line below to match `filter()`'s new position.

mlehtima: Fixed

PR #138, opened 2023-02-20
[sailjail-permissions] Remove disable-passwdmgr.inc. Contributes to JB#59121

dsuni:
Firejail 0.9.72 no longer uses or includes disable-passwdmgr.inc. Having it present in sailjail-permissions causes sailjail to fail to launch with the newer firejail.

### DIFF
--- a/permissions/Base.permission
+++ b/permissions/Base.permission
@@ -26,7 +26,6 @@ include disable-common.inc
 # include disable-devel.inc
 # include disable-exec.inc
 # include disable-interpreters.inc
-include disable-passwdmgr.inc
 # include disable-programs.inc
 # include disable-xdg.inc
 
--- a/rpm/sailjail-permissions.spec
+++ b/rpm/sailjail-permissions.spec
@@ -1,6 +1,6 @@
 Name:     sailjail-permissions
 Summary:  Permissions definitions for Sailjail
-Version:  1.0.0
+Version:  1.1.10
 Release:  1
 License:  BSD
 URL:      https://github.com/sailfishos/sailjail-permissions


permissions/Base.permission
```diff
@@ -26,7 +26,7 @@ include disable-common.inc
 # include disable-devel.inc
 # include disable-exec.inc
 # include disable-interpreters.inc
-include disable-passwdmgr.inc
+# include disable-passwdmgr.inc
```
pvuorela, line 29: If the file doesn't exist anymore on the new version, just remove the whole reference to it.

rpm/sailjail-permissions.spec
```diff
@@ -1,6 +1,6 @@
 Name:     sailjail-permissions
 Summary:  Permissions definitions for Sailjail
-Version:  1.0.0
+Version:  1.0.1
```
pvuorela, line 3: If you touch this, it should be the roughly the new to be tagged version, i.e. 1.1.10.

pvuorela (approved):
Approving.

For final nitpick the commit message summary:
[sailjail-permissions] Remove disable-passwdmgr.inc Facilitates JB#59121 

Instead of "facilitates" we've used "Contributes to JB#...". Those words are not used for anything special but good to follow the practice so things don't unnecessarily stand out in the history. Also usually period before the bug reference part.

Tomin1:
> @Tomin1 ponder should we just merge this already to pave way for the update? The .inc file doesn't appear such that it should have a drastic effect whether included or not.

Yes, sure. I agree.

PR #139, opened 2023-04-18
[permissions] Document Privileged permission. JB#60573

vigejolla:


martyone (approved):
Maybe add a link to related documentation (if we have any)?

vigejolla:
> Maybe add a link to related documentation (if we have any)?

I'm not aware of any such documentation. We should write some.

PR #140, opened 2023-04-21
[permissions] Move Privileged permission to internal list. JB#60573

vigejolla:
The Privileged permission should be part of other permissions where needed. In general there shouldn't be reasons for applications to use Privileged permission directly.


Tomin1 (approved):
:+1:

PR #141, opened 2023-10-10
[sailjail-permissions] Move Email permission to DBus.

dcaliste:
The message server is not using a specific
socket file, but is now using DBus.

@pvuorela, this could be usefull to test the QMF switch to DBus on device.

### DIFF
--- a/permissions/Email.permission
+++ b/permissions/Email.permission
@@ -6,10 +6,11 @@
 # x-sailjail-translation-key-long-description = permission-la-email_description
 # x-sailjail-long-description = Read and send email
 
-# MessageServer socket
-mkdir     ${RUNUSER}/messageserver
-whitelist ${RUNUSER}/messageserver
-read-only ${RUNUSER}/messageserver
+# MessageServer DBus interfaces
+dbus-user.talk org.qt.messageserver
+dbus-user.talk org.qt.mailstore
+dbus-user.broadcast org.qt.messageserver=org.qt.messageserver.*@/*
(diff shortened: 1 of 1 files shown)


permissions/Email.permission
```diff
@@ -6,10 +6,11 @@
...
 
-# MessageServer socket
-mkdir     ${RUNUSER}/messageserver
-whitelist ${RUNUSER}/messageserver
-read-only ${RUNUSER}/messageserver
+# MessageServer DBus interfaces
+dbus-user.talk org.qt.messageserver
+dbus-user.talk org.qt.mailstore
+dbus-user.broadcast org.qt.messageserver=org.qt.messageserver.*@/messageserver
+dbus-user.broadcast org.qt.mailstore=org.qt.mailstore.*@/mailstore/client
```
pvuorela, line 13: Think we could allow everything behind the interface, i.e. something like @/*. Per se the clients are free to choose whatever path for their signalling objects and hard-coding the current one could have a risk of silently breaking if that ever gets adjusted. Shouldn't bring either extra security since only the specific interface is allowed.

The server is more of a well-known object path, but even there could just allow everything, as seems commonly done on other rules. That would also work if the action api is adjusted to use separate objects.

dcaliste: I see your point. Not changing this file at each minor modification / addition would be nicer, keeping the same degree of protection.

I moved to a `/*` filter for paths.

pvuorela:
Seems good, though need to wait for other changes in first.

pvuorela (approved):

PR #144, opened 2024-09-25
[sailjail-permissions] Allow to listen to commhistory database modifi…

dcaliste:
…cations.

Using the modification in libcommhistory, this whitelist signals emitted by any process ;odifying the comm history database.

### DIFF
--- a/permissions/CommunicationHistory.permission
+++ b/permissions/CommunicationHistory.permission
@@ -12,3 +12,6 @@ whitelist ${HOME}/.local/share/commhistory
 dbus-user.talk org.nemomobile.CommHistory
 dbus-user.broadcast org.nemomobile.CommHistory=org.nemomobile.CommHistory.*@/org/nemomobile/CommHistory
 dbus-user.call org.nemomobile.CommHistory=org.nemomobile.CommHistory.*@/org/nemomobile/CommHistory
+
+dbus-user.own com.nokia.commhistory.*
+dbus-user.broadcast *=com.nokia.commhistory.*@/CommHistoryModel


pvuorela:
Hm, there are some other rules with * for service, e.g.
`dbus-user.broadcast *=org.nemomobile.contacts.sqlite.*@/org/nemomobile/contacts/sqlite`

Wonder if such would work here.

dcaliste:
It's working because qtcontact is using the same solution : the library declare itself as a service named org.nemomobile.contact.sqlite.uuid_PID, see contactnotifier.cpp. So the broadcast can use a '*' for service name.

I can also put a * for the broadcast, but it's better in my opinion to restrict to the name pattern the library is using.

Besides, I should do the same than for Contacts : owning the name service and not talk to.

dcaliste:
If you look at the xdg_dbus_proxy sources, in flatpack-proxy.c line 2475, you can notice that the broadcast whitelist filtering is done only if the policy is OWN or TALK. And to talk or own, it must be a named service.

pvuorela:
> If you look at the xdg_dbus_proxy sources, in flatpack-proxy.c line 2475, you can notice that the broadcast whitelist filtering is done only if the policy is OWN or TALK. And to talk or own, it must be a named service. 

Was actually checking that repo a while ago. On latest release there was a commit "Don't require TALK permission for broadcast rules". Tried an update, but there's one commit conflicting. Upstream took a bit different route but didn't yet get a conclusion if it's sufficient for us. I might try to get the update proceeding at some point now.

> I can also put a * for the broadcast, but it's better in my opinion to restrict to the name pattern the library is using.

I think it's better the less we make artificial restrictions. The D-Bus API wouldn't otherwise need template based service name. It shouldn't be much of a security hole either to be able to get signals on specific interface regardless of what services their sender has.

dcaliste:
> I think it's better the less we make artificial restrictions.

Ok, I've put a wildcard for the service name the interface is originating from.

> On latest release there was a commit "Don't require TALK permission for broadcast rules".

That's interesting. Could be part of the solution. But thinking more about it and looking at the dbus-user.own of the contact case, I think we need the own rule with the name template anyway. In order for the processes outside the jail to get notified of a change that may be done by the application inside the jail (like when deleting one entry of the call log by long press for instance). There is no usecase at the moment, since no other application is listing the call log. But to avoid such issue as the current one in the future, it's better to allow to own the service name and thus also emit signals from inside. What do you think ?

pvuorela:
That's a fair point on being able to get the signals outside the sandbox. And especially when there's similar use for contacts.

pvuorela (approved):

PR #150, opened 2026-02-11
[sailjail-permissions] Allow access to last position for qtlocation

nephros:
If I understand correctly, the Qt GeoClue plugin stores its last known position in a per-user data file.

Allow reading and writing that file in the Location permission.

For context, see

 - https://github.com/sailfishos/qtlocation/blob/a1c218a04e8e56206fbbbbb8f071eb33484e80a3/src/plugins/position/geoclue/qgeopositioninfosource_geocluemaster.cpp#L82
 - https://github.com/sailfishos/qtlocation/blob/a1c218a04e8e56206fbbbbb8f071eb33484e80a3/src/plugins/position/geoclue/qgeopositioninfosource_geocluemaster.cpp#L104

### DIFF
--- a/permissions/Location.permission
+++ b/permissions/Location.permission
@@ -11,6 +11,9 @@ whitelist /var/lib/location
 
 private-etc location
 
+# allow retrieving and storing last known position
+read-write ${HOME}/.local/share/qtposition-geoclue
+
 include /etc/sailjail/permissions/Sensors.permission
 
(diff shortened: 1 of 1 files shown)


pvuorela:
Code is there in qtlocation but it's a bit suspicious too if apps are able to write such a thing (though if the don't, then no-one is going to). Also the implementation reads it no matter how old. Pondering should geoclue do something like this instead, though not entirely unproblematic even there.

